PR #1, opened 2017-03-19
Coloured reporting and Template Haskell runner

jystic:
This significantly improves the test runner, adding coloured output 🌈 and some template haskell code for discovering properties.

screen shot 2017-03-19 at 5 33 44 pm

The property discovery is a little hacky, but it does deal with multi-line comments, so the following will not be picked up as a property:
```hs
{-
prop_foo =
  not done yet
-}
```

The failure output pulls in the source code of the property and annotates it with the generated test data, as well as "calling out" which of your propositions actually failed. I can imagine this will be useful when you have a `conjoin` like thing with many possible ways in which you can fail:

screen shot 2017-03-19 at 5 30 21 pm

The source code annotation relies on the GHC 8.0 `CallStack` API, so I'm going to need to figure out something to do as a stop gap for use in GHC 7.10. I'll probably just go back to printing the forall's in a list along with their types if compiling for 7.10.

Ignore the churn on the `Example.hs` file, I was just playing with things to see what different failures would look like in the test runner. Hopefully the source code annotation stuff scales to larger examples, I'm not sure, but I think it should be ok.

My todo list before doing the first Hackage release is basically just adding examples for working with `IO`, `ResourceT` and `ExceptT` to see if there's any issues with Template Haskell property discovery stuff and the interface for monadic tests.

### DIFF
--- a/hedgehog.cabal
+++ b/hedgehog.cabal
@@ -34,15 +34,17 @@ source-repository head
 library
   build-depends:
       base                            >= 3          && < 5
-    , base64-bytestring               >= 1.0        && < 1.1
+    , ansi-terminal                   >= 0.6        && < 0.7
     , bytestring                      >= 0.10       && < 0.11
     , containers                      >= 0.4        && < 0.6
     , mmorph                          >= 1.0        && < 1.1
     , pretty-show                     >= 1.6        && < 1.7
     , random                          >= 1.1        && < 1.2
+    , template-haskell                >= 2.10       && < 2.12
     , text                            >= 1.1        && < 1.3
     , time                            >= 1.4        && < 1.9
     , transformers                    >= 0.3        && < 0.6
+    , wl-pprint-annotated             >= 0.0        && < 0.1
 
   if impl(ghc < 8.0)
     build-depends:
@@ -55,10 +57,15 @@ library
     src
 
   exposed-modules:
+    Hedgehog.Example
     Hedgehog.Gen
     Hedgehog.Property
     Hedgehog.Range
+    Hedgehog.Runner
+    Hedgehog.TH
 
+    Hedgehog.Internal.Discovery
+    Hedgehog.Internal.Report
     Hedgehog.Internal.Seed
     Hedgehog.Internal.Shrink
     Hedgehog.Internal.Tree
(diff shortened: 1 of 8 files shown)


hedgehog.cabal
```diff
@@ -34,15 +34,17 @@ source-repository head
 library
   build-depends:
       base                            >= 3          && < 5
-    , base64-bytestring               >= 1.0        && < 1.1
+    , ansi-terminal                   >= 0.6        && < 0.7
```
charleso: I know it would be less convenient, but I wonder whether it would be worth splitting up the project into core, runner-terminal, example, etc. I suspect that will be too annoying for the first release.

charleso:
Updated screenshots or output would be good in the doc. :)

charleso:
Next level. 💯

hedgehog.cabal
jystic: Yeah I think it's probably worth waiting until there is a concrete burden before doing this

jystic: I'm not too worried about `ansi-terminal` as a dependency, it's absolutely micro and only depends libraries which already ship with GHC itself (`unix` / `Win32` depending on the platform).

src/Hedgehog/Internal/Report.hs
```diff
@@ -0,0 +1,611 @@
...
+  -- FIXME This is pretty savage, not sure how else to provide this while
+  -- FIXME maintain a seamless experience. Maybe we should just go for a
+  -- FIXME HEDGEHOG_COLOR environment variable instead.
+  args <- liftIO getArgs
+  if elem "--color" args then
+    pure True
+  else if elem "--no-color" args then
+    pure False
+  else
+    liftIO $ hSupportsANSI stdout
```
thumphries: Agree with your env variable suggestion, this `getArgs` is pretty brutal.

`hSupportANSI` will be correct for 99% of users though 👍

thumphries:
this is amazing 🔥

src/Hedgehog/Internal/Report.hs
jystic: unfortunately it doesn't work with `cabal test` (and hence `mafia test`) !! :(

jystic: so the only way to get colours is to say `mafia test -- --test-option=--color`, but yeah an env var would do the trick too, works fine in `ghci`

PR #4, opened 2017-03-19
Hutton's razor example

jystic:
The inline output works out pretty well for this example:

screen shot 2017-03-19 at 10 05 16 pm


charleso:
👍

PR #8, opened 2017-03-21
Add top level Hedgehog module

jystic:
This adds a top level module so people don't need to hunt for imports, it also has some placeholder doc which I will elaborate on before release.

I've also twiddled the names, gone from `ensure` back to `assert` and `forAll` is now `given`. I'd be interested in any feedback on this. I'm not particularly concerned with being familiar to QuickCheck users anymore given how different this library is now, so I'm mainly interested in how you think properties read using these names for things.

### DIFF
--- a/hedgehog-example/test/Test/Example/Basic.hs
+++ b/hedgehog-example/test/Test/Example/Basic.hs
@@ -6,11 +6,9 @@ module Test.Example.Basic where
 
 import           Control.Monad (guard)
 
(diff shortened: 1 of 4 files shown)


thumphries:
`assert` was good. I was pretty confused when it became `ensure`.

I preferred `forAll` to `given` though. QuickCheck doesn't own that word!

jystic:
Ok, you've convinced me for now, will change it back to `forAll`, I just didn't like that it was two words

PR #9, opened 2017-03-21
Implement mtl + other common monad type classes

jystic:
My brain hurts

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Tree.hs
+++ b/hedgehog/src/Hedgehog/Internal/Tree.hs
@@ -1,4 +1,9 @@
 {-# LANGUAGE CPP #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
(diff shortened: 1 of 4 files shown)


hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -136,18 +146,134 @@ instance MonadTrans Tree where
...
-      Tree . f $ fmap hoistOutcome m
+    Tree . f $ fmap (hoist f) m
+
+embedNode :: Monad m => (t (Node t b) -> Tree m (Node t b)) -> Node t b -> Node m b
+embedNode f (Node x xs) =
+  Node x (fmap (embedTree f) xs)
+
+embedTree :: Monad m => (t (Node t b) -> Tree m (Node t b)) -> Tree t b -> Tree m b
+embedTree f (Tree m) =
+  Tree . pure . embedNode f =<< f m
```
tranma: wow

thumphries:
Thanks, this makes it a lot easier to use.

PR #11, opened 2017-03-22
WIP STLC example

thumphries:
This could use a fair bit of cleaning up, but is mostly there. The generators are working reasonably well, though they may not be idiomatic. `genWellTypedApp` is producing reasonably interesting chains of function application.

As you can see, I haven't bothered implementing `subst`, should probably do that.

@jystic @charleso

### DIFF
--- a/hedgehog-example/test/Test/Example/STLC.hs
+++ b/hedgehog-example/test/Test/Example/STLC.hs
@@ -1,0 +1,258 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+module Test.Example.STLC where
+
+
+import           Control.Applicative
+import           Control.Monad
+import           Control.Monad.Morph
+import           Control.Monad.Reader
+
+import           Data.Either
+import           Data.Map.Strict (Map)
+import qualified Data.Map.Strict as M
+import           Data.Maybe
+import           Data.Monoid
+import           Data.Set (Set)
+import qualified Data.Set as S
+import           Data.Text (Text)
+
+import           Hedgehog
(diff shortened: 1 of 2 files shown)


hedgehog-example/test/Test/Example/STLC.hs
```diff
@@ -0,0 +1,281 @@
...
+  | EInt Int
+  | EString Text
+  | EVar Text
+  | ELam Text Type Expr
+  | EApp Expr Expr
+  deriving (Eq, Ord, Show)
+
+-- -----------------------------------------------------------------------------
+
+-- | Evaluate to weak head normal form.
```
thumphries: would probably get more interesting failures if this went all the way to normal form, was just being lazy.

thumphries:
Not sure how to minimise this so it's a good teaching example, just threw in most of the techniques I've found useful.

hedgehog-example/test/Test/Example/STLC.hs
```diff
@@ -0,0 +1,281 @@
...
+  eg <- genWellTypedExpr' tg
+  let tf = TArrow tg want
+  ef <- genWellTypedExpr' tf
+  pure (EApp ef eg)
+
+-- | This tries to look up a known expression of the desired type from the env.
+-- It does not always succeed, throwing `empty` when unavailable.
+genWellTypedPath :: Type -> Gen (Reader (Map Type [Expr])) Expr
+genWellTypedPath want = do
+  paths <- lift ask
```
jystic: You should be able to replace this with just `ask` now we have `MonadReader`

jystic:
I'm going to merge this so I can keep it up to date with my churning

jystic:
I think there's no harm having it in the examples area, happy for it to be cleaned up whenever you have time.

PR #12, opened 2017-04-02
Clean up / polishing

jystic:
Not sure if I really like the split I've created here between Property / Test, but I can't think of a better way to allow control over the number of tests / discards / shrinks.

The split allows stuff like this:
```hs
prop_once :: Property
prop_once =
  withTests 1 . property $ do
    success

prop_lots :: Property
prop_lots =
  withTests 10000 . withDiscards 10000 . property $ do
    success

prop_no_shrink :: Property
prop_no_shrink =
  withShrinks 0 . property $ do
    success
```

Theres a lot of changes here, mainly driven by trying to write examples, certainly interested in any thoughts you have about the Property / Test split.

Some highlights:
- Example using `ResourceT` and `ExceptT`.
- Exceptions that occur during a test are now caught, but this isn't working very nicely yet, the exception cuts off the `WriterT` so we don't see the input data we generated that caused the error.
- Added support for shrinking to sub-terms, see `subterm` / `subtermM`
- Added docs, I think all the public functions now have haddock

The way `ResourceT` interacts with shrinking is workable, but not very nice, unfortunately I can't see a better solution at the moment.

/cc @thumphries @charleso


hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -0,0 +1,369 @@
...
+import           Text.Show.Pretty (ppShow)
+
+------------------------------------------------------------------------
+
+-- | A property test to check, along with some configurable limits like how
+--   many times to run the test.
+--
+data Property =
+  Property {
+      propertyConfig :: !Config
```
charleso: 💯   to having this per property

charleso:
👍 

I think the split makes sense. The implications of side-effecting `m` + shrinking seems interesting, and we'll have to wait to see how it plays out

PR #13, opened 2017-04-03
Add diffs when using the === operator

jystic:
Not sure if the look/feel is quite right yet, but it's a start

screen shot 2017-04-04 at 8 02 37 am

screen shot 2017-04-04 at 8 12 45 am


moodmosaic:
This looks really cool!

PR #14, opened 2017-04-07
Adds `Read` instances for `Size` and `Seed`.

dalaing:
These instances should work with the existing `Show` instances.

I've added them as a precursor to work on `tasty-hedgehog`.


tasty-hedgehog/src/Test/Tasty/Hedgehog.hs
```diff
@@ -0,0 +1,125 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
moodmosaic: Hmm... Wouldn't it be more flexible if `tasty-hedgehog` (and any other testing framework implementation) evolves outside of the main `haskell-hedgehog` repository?

I actually liked the simplicity of the approach shown [here](https://github.com/hedgehogqa/haskell-hedgehog/blob/aed5c8892ce5a955f425a1e31ce7b452ed5376c1/hedgehog/src/Hedgehog.hs#L23-L28), where you can just add a Template Haskell splice to run the properies.

(I'm not saying that the proposed `tasty-hedgehog` is a bad idea, just that it can be easier if its development happens outside of the main repository.)

dalaing: Whoops - I didn't mean to include that in this branch, this PR was only meant to be about the Read instances. I'll clean that up shortly.

dalaing:
I'm killing this PR, and will replace it with one not afflicted by late night git autopilot.

PR #15, opened 2017-04-07
Adds `Read` instances for `Size` and `Seed`.

dalaing:
These should match the `Show` instances.

### DIFF
--- a/hedgehog/test/Text.hs
+++ b/hedgehog/test/Text.hs
@@ -2,7 +2,6 @@ module Text (
     textProperties
   ) where
 
-import           Data.Foldable (elem)
 import           Data.Typeable (Typeable)
 
 import           Hedgehog
--- a/hedgehog/test/Text.hs
+++ b/hedgehog/test/Text.hs
@@ -1,0 +1,62 @@
+module Text (
+    textProperties
+  ) where
+
+import           Data.Foldable (elem)
+import           Data.Typeable (Typeable)
+
+import           Hedgehog
+import qualified Hedgehog.Gen as G
+import qualified Hedgehog.Range as R
+
+genSize :: Monad m
+        => G.Gen m Size
+genSize =
+  Size <$> G.enumBounded
+
+genSeed :: Monad m
+        => G.Gen m Seed
+genSeed =
+  Seed <$> G.enumBounded <*> oddGen
+    where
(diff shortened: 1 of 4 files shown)


charleso:
I know testing has been thin on the ground while Jacob has been smashing things out, but should we add a round-trip read/show property test for these? :)

dalaing:
That's a very good call.

Also, while I'm at it: I'm guessing I should write the `Read` instances that correspond to the other `Show` instances in `hedgehog`, rather than just doing what I need and floating on to the next thing :)

jystic:
> Also, while I'm at it: I'm guessing I should write the Read instances that correspond to the other Show instances in hedgehog, rather than just doing what I need and floating on to the next thing :)

Only if you actually need them, I'd be ok with only having `Read` for `Size` and `Seed`, the other stuff probably won't end up getting rendered anywhere.

hedgehog/test/Text.hs
```diff
@@ -0,0 +1,62 @@
...
+  a <- forAll g
+  elem (a, "") (readsPrec p (showsPrec p a "")) === True
+
+textProperties :: [Property]
+textProperties =
+  [ showAppendProp genSize
+  , readShowProp genSize
+  , showAppendProp genSeed
+  , readShowProp genSeed
+  ]
```
jystic: You might as well use `$((checkAll))` here (just need to name the properties starting with `prop_`)

jystic: oh sorry, I see what you've done, yeah fair enough

hedgehog/test/Text.hs
```diff
@@ -0,0 +1,62 @@
+module Text (
+    textProperties
+  ) where
+
+import           Data.Foldable (elem)
```
jystic: I don't think this is required, seems to be breaking the build for 7.10:
```
test/Text.hs:5:1: Warning:
    The import of ‘Data.Foldable’ is redundant
      except perhaps to import instances from ‘Data.Foldable’
    To import instances alone, use: import Data.Foldable()
```

jystic:
Sorry to be a pain, I haven't got a build set up to check this yet, but would you mind rebasing this on top the latest master? Github doesn't have a rebase + merge --no-ff button which is how I want this repo work.

jystic:
Have rebased and merged

PR #17, opened 2017-04-08
Concurrent test runner

jystic:
This adds support for running tests concurrently.

screen shot 2017-04-08 at 7 29 30 pm


thumphries:
This is awesome.

Does concurrent-output do the right thing when noninteractive?

jystic:
It does with a bit of help, see `flushRegions` and `finishIndexedRegion`

When you "finish" a region, that is what is printed in non-tty mode. The issue is that if you finish a region when a test actually completes then you get a somewhat random order of output. So those two functions make sure that we never finish a region before the tests which are ordered prior to it have also been finished. It's probably a bit pendantic, but it makes the whole thing feel a lot nicer.

hedgehog/src/Hedgehog/Internal/TH.hs
```diff
@@ -21,14 +21,38 @@ import           Language.Haskell.TH.Syntax
...
   Q (TExp a)
 
--- | Check all the properties in a file.
+-- | Check all the properties in a file sequentially.
+--
+-- > tests :: IO Bool
+-- > tests =
+-- >   $$(checkSequential)
+--
+checkSequential :: TExpQ (IO Bool)
```
charleso: I should have mentioned this before, but I often wondered whether it would be better to generate just the bare minimum, and then do the "running" at the top level. So rather than `IO Bool`, just generate the bare minimum `Map Name Source`. That said, you may be planning to have files that are either sequential and some concurrent.

jystic: I guess you'd be able to better distribute the load that way, but we'd need a way to mark a test a concurrent or sequential, maybe per property is a better approach anyway

PR #27, opened 2017-04-09
Export reporting constructors

erikd:


hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -11,9 +11,9 @@ module Hedgehog.Internal.Report (
   , FailureReport(..)
   , FailedInput(..)
 
-  , ShrinkCount
-  , TestCount
-  , DiscardCount
+  , ShrinkCount(..)
+  , TestCount(..)
+  , DiscardCount(..)
```
jystic: These are all `Num` + `Integral` already, so you don't need the constructor exposed to get at them if you want it as an `Int` (just use `fromIntegral`)

erikd: Fair enough.

jystic: I don't necessarily have a problem exposing the constructors, they're not protecting anything special

PR #31, opened 2017-04-09
Runner: Add checkReportNoCatch

erikd:
Need this for Hspec integration. Hspec has the ability to check that
the correct exception gets thrown from a test and hence needs the
exception to escape the test runner.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -12,6 +12,7 @@ module Hedgehog.Internal.Runner (
 
   -- * Internal
   , checkReport
+  , checkReportNoCatch
   , checkConsoleRegion
   , checkNamed
   ) where
@@ -126,11 +127,20 @@ checkReport ::
   -> Test m ()
   -> (Report -> m ())
   -> m Report
-checkReport cfg size0 seed0 test0 updateUI =
-  let
-    test =
-      catchAll test0 (fail . show)
+checkReport cfg size0 seed0 test updateUI =
+  checkReportNoCatch cfg size0 seed0 (catchAll test (fail . show)) updateUI
 
+checkReportNoCatch ::
+     forall m.
+     MonadIO m
+  => PropertyConfig
+  -> Size
+  -> Seed
+  -> Test m ()
+  -> (Report -> m ())
+  -> m Report
+checkReportNoCatch cfg size0 seed0 test updateUI =
+  let
     loop :: TestCount -> DiscardCount -> Size -> Seed -> m Report
     loop !tests !discards !size !seed = do
       updateUI $ Report tests discards Running


dalaing:
Perhaps slightly orthogonal to this PR, but it might be nice to have a `ReportConfig` type in order to do modifications like this to the report generation.  

For `tasty-hedgehog`, it'd be nice to be able to suppress the instructions on how to reproduce the failure.

I can see that there would be some scope for different levels of verbosity as well - no details of the failure, the current level of details, the current level + details of discards and shrinks, etc...

At least being able to suppress or condense the failure details would help to achieve parity with `tasty-quickcheck`.

erikd:
@dalaing Yeah, there's a couple of small changes that might be required for integration with things like `tasty` and `hspec`.

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -126,11 +127,20 @@ checkReport ::
...
+checkReportNoCatch ::
+     forall m.
+     MonadIO m
+  => PropertyConfig
+  -> Size
+  -> Seed
+  -> Test m ()
+  -> (Report -> m ())
+  -> m Report
+checkReportNoCatch cfg size0 seed0 test updateUI =
```
jystic: I don't think this will do what you want, you'll end up checking that any of the 100 or so tests throws an exception, but it sounds like you want to check that all of the tests throw an exception

erikd: > it sounds like you want to check that all of the tests throw an exception.

Yes, thats what I want. Suggestions?

jystic: Hmm, it's like you want the test loop to be in hspec instead? The fact that every property runs 100s of tests makes this problematic. What does hspec do for QuickCheck?

jystic: I think I need to have a look at how hspec works, but it seems a little odd that it does more things than just running the tests? Surely it's the job of the test itself to pass/fail?

erikd: Here's an [example](https://github.com/erikd/hspec-hedgehog/blob/master/test/Test/Test/Hspec/Hedgehog.hs#L29-L30):
```
      Hspec.it "propagates exceptions" $ do
        evaluateExample (error "foobar") `shouldThrow` errorCall "foobar"
```

jystic: What happens if you do the same thing with the hspec quickcheck integration?

erikd: A simple Hspec with QuickCheck program to test this:
```
{-# LANGUAGE ScopedTypeVariables #-}
import           Data.Word
import           Test.Hspec
import           Test.Hspec.QuickCheck

main :: IO ()
main = hspec $
  prop "exception handling" $ \ (i :: Word8) -> 
    throwError i `shouldThrow` errorCall ("error " ++ show i)

throwError :: Word8 -> IO a
throwError i = do
  let msg = "error " ++ show i
  putStrLn msg
  error msg
```
shows that `throwError` does indeed get called more than once.

Its actually going to be the same for Hedgehog because its the `Test.Hspec.QuickCheck.prop` function that makes sure that the condition is evaluated more than once. I just need to implement my own `Test.Hspec.Hedgehog.prop` function.

So it seems that what Hedgehog needs to expose is some way of allowing a property to be tested just once, but with a test state that is maintained over multiple calls.

jystic: > So it seems that what Hedgehog needs to expose is some way of allowing a property to be tested just once, but with a test state that is maintained over multiple calls.

I think this makes sense.

jystic:
I'm going to close this PR because I don't think it makes sense to let exceptions leak out of the main test loop.

> So it seems that what Hedgehog needs to expose is some way of allowing a property to be tested just once, but with a test state that is maintained over multiple calls.

^ From previous discussion it sounds like this it what you really want.

erikd:
> So it seems that what Hedgehog needs to expose is some way of allowing a property to be tested just once, but with a test state that is maintained over multiple calls.

Should I raise a ticket for that?

jystic:
Yeah good idea 👍 

Maybe include the details of how hspec works with QuickCheck which explained above?

erikd:
#84

PR #35, opened 2017-04-10
Make it easier to use Hedgehog without Template Haskell

jystic:
This adds a type `Group` which is a named collection of properties.

`checkConcurrent` and `checkSequential` are now regular functions which can execute the property tests in a `Group`. The Template Haskell property discovery is now just a single function `discover` which returns a `Group`.

Running tests with automatic discovery now looks like this:

```hs
tests :: IO Bool
tests =
  checkConcurrent $$(discover)
```

But now, if you prefer to avoid macros, you can specify the group of properties to run manually instead:

```hs
tests :: IO Bool
tests =
  checkConcurrent $ Group "Test.Example" [
      ("prop_reverse", prop_reverse)
    ]
```

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/TH.hs
+++ b/hedgehog/src/Hedgehog/Internal/TH.hs
@@ -2,9 +2,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 module Hedgehog.Internal.TH (
     TExpQ
-  , checkSequential
-  , checkConcurrent
-  , checkWith
+  , discover
   ) where
 
 import qualified Data.List as List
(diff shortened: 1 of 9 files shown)


hedgehog/src/Hedgehog/Internal/TH.hs
```diff
@@ -13,46 +11,19 @@ import qualified Data.Ord as Ord
...
-    RunnerConfig {
-        runnerWorkers =
-          Nothing
-      }
-
--- | Check all the properties in a file.
---
-checkWith :: RunnerConfig -> TExpQ (IO Bool)
-checkWith config = do
+discover :: TExpQ Group
```
charleso: 💯  :)

charleso:
Nice 👍

moodmosaic:
Nice 👍

Issue #34, opened 2017-04-10
Does it make sense to have a Gen for ()?


newhoggy:

jystic:
As in `Gen.unit :: Gen ()` ?

Perhaps for consistency? but the implementation can only be `pure ()` I suppose?

newhoggy:
That works.

I had been using:

    unit :: Monad m => Gen.Gen m ()
    unit = Gen.enumBounded

But `pure ()` is better.

I wrote a semigroup property test for pedagogical purposes.  😁

chris-martin:
I also found myself needing this and didn't realize it could be done with pure. Suggested #36 as a documentation fix.

jystic:
@newhoggy are you ok to close out this issue with the addition of the new `Gen.constant` combinator?

newhoggy:
`pure` works for me.  Thanks!

PR #36, opened 2017-04-11
add Gen.constant

chris-martin:
I spent a while looking for this function, and eventually found #34 and felt silly when I realized you can just use `pure` for this. So I figured since at least two of us have made this mistake so far, it'd be good to leave some kind of reminder in the documentation.

I propose adding an alias called `constant`, with a note explaining that it is equivalent to `pure`.

It didn't quite fit neatly into any of the sections, but I put it under "Choice" because you can think of it as a specialization of `element` with a single value, which I also point out in the documentation.

### DIFF
--- a/hedgehog/src/Hedgehog/Gen.hs
+++ b/hedgehog/src/Hedgehog/Gen.hs
@@ -74,6 +74,7 @@ module Hedgehog.Gen (
   , bytes
 
   -- ** Choice
+  , constant
   , element
   , choice
   , frequency
@@ -767,6 +768,13 @@ bytes range =
 
 ------------------------------------------------------------------------
 -- Combinators - Choice
+
+-- | Trivial generator that always produces the same element.
+--
+--   /This is another name for 'pure' \/ 'return'./
+constant :: Monad m => a -> Gen m a
+constant =
+  pure
 
 -- | Randomly selects one of the elements in the list.
 --


hedgehog/src/Hedgehog/Gen.hs
```diff
@@ -768,6 +769,16 @@ bytes range =
 ------------------------------------------------------------------------
 -- Combinators - Choice
 
+-- | Trivial generator that always produces the same element.
+--
+-- All of the following expressions are equivalent:
+--
+-- * @'constant' a@
+-- * @'pure' a@
+-- * @'element' [a]@
```
jystic: For `discard` which serves a similar purpose, I have used the words `This is another name for 'empty' / 'mzero'.`

So just for consistency I'd prefer this to say:
```
-- | Trivial generator that always produces the same element.
--
--   /This is another name for 'pure' \/ 'return'./
```

hedgehog/src/Hedgehog/Gen.hs
```diff
@@ -768,6 +769,16 @@ bytes range =
...
+constant :: Monad m => a -> Gen m a
+constant = pure
```
jystic: I should write up the style I'm following in this repo so I can just point to it, but this needs to be:
```hs
constant :: Monad m => a -> Gen m a
constant =
  pure
```

jystic:
Happy with this in general, we have the [same combinator](https://github.com/hedgehogqa/fsharp-hedgehog/blob/a254aed9c61bec168d31f58dca00d62186ecb7d7/src/Hedgehog/Gen.fs#L29-L30) in the F# version, just a few nitpicks.

hedgehog/src/Hedgehog/Gen.hs
chris-martin: Do you care what we do with the comparison to `element`?

hedgehog/src/Hedgehog/Gen.hs
chris-martin: Ah yes, I should have noticed that.

chris-martin:
Alright, changes made.

hedgehog/src/Hedgehog/Gen.hs
jystic: I think your primary description is clear enough to be honest :)

jystic:
👍 cool, looks good

PR #42, opened 2017-04-11
Make the examples runnable with doctest

chris-martin:
With these couple of tweaks, the code examples can be made runnable with [doctest](https://hackage.haskell.org/package/doctest).

```bash
> doctest hedgehog/src/Hedgehog
Examples: 34  Tried: 34  Errors: 0  Failures: 0
```

This PR also fixes an error in one of the examples, a `towards` that should be `towardsFloat` (a problem that I found by running doctest :smile:)

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Shrink.hs
+++ b/hedgehog/src/Hedgehog/Internal/Shrink.hs
@@ -37,11 +37,11 @@ towards destination x =
 
 -- | Shrink a floating-point number by edging towards a destination.
 --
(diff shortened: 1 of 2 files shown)


charleso:
Just an observation from the sideline but if we want to do this is it worth adding doctest to the build in some way so we pick up these things errors rather than let them get out of date?

jystic:
I think we do, but I'm happy to take care of the build stuff myself in a later PR.

jystic:
👍 have rebased and merged

PR #43, opened 2017-04-11
Add exponential ranges

### LINKED ISSUE #37: Exponential Range
Is the plan for `Hedgehog.Range` that it will include variations other than constant and linear? In particular I think being able to generate an integer that scales exponentially with the size parameter would make sense in some cases.

chris-martin:
Closes #37

### DIFF
--- a/hedgehog/src/Hedgehog/Range.hs
+++ b/hedgehog/src/Hedgehog/Range.hs
@@ -23,11 +23,20 @@ module Hedgehog.Range (
   , linearFracFrom
   , linearBounded
 
+  -- * Exponential
+  , exponential
+  , exponentialFrom
+  , exponentialBounded
+  , exponentialFloat
+  , exponentialFloatFrom
+
   -- * Internal
   -- $internal
   , clamp
   , scaleLinear
   , scaleLinearFrac
+  , scaleExponential
+  , scaleExponentialFloat
   ) where
 
 import           Data.Bifunctor (bimap)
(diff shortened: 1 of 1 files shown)


hedgehog/src/Hedgehog/Range.hs
```diff
@@ -296,6 +305,148 @@ scaleLinearFrac sz0 z n =
...
+-- | Scale a floating-point number exponentially with the size parameter.
+--
+scaleExponentialFloat :: Floating a => Size -> a -> a -> a
+scaleExponentialFloat sz0 z n =
+  let
+    sz =
+      clamp 0 99 sz0
+
+    diff =
+      (((abs (n - z) + 1) ** (realToFrac sz / 99)) - 1) * signum (n - z)
```
chris-martin: The outcome seems right, but my math could probably use a double-check here. I seem to be struggling to recall my high school algebra today.

jystic:
I should get a chance to look at this tonight

chris-martin:
Some notes on the design:

* The way I'm doing the calculation involves a non-integral exponent, so I had to use floating-point arithmetic.
* The `Integral` functions are just based on the `Floating` functions, with a rounding step at the end.
* We *don't* have `Fractional` exponential ranges; I'm not sure whether there is a reasonable way to do that.

hedgehog/src/Hedgehog/Range.hs
```diff
@@ -300,6 +309,148 @@ scaleLinearFrac sz0 z n =
...
+--   (-2,4)
+--
+--   >>> bounds 50 $ exponentialFrom 0 (-128) 512
+--   (-11,22)
+--
+--   >>> bounds 75 $ exponentialFrom 0 (-128) 512
+--   (-39,112)
+--
+--   >>> bounds 99 $ exponentialFrom 0 (-128) 512
+--   (-128,512)
```
jystic: These examples should probably have `x` instead of `0` because those things are true for all `x`

hedgehog/src/Hedgehog/Range.hs
jystic: Maybe we could change this to:
```hs
scaleExponentialFrac :: RealFrac a => Size -> a -> a -> a
scaleExponentialFrac sz0 z n =
  let
    sz =
      clamp 0 99 sz0

    p :: Double
    p =
      realToFrac (abs (n - z) + 1) ** (realToFrac sz / 99)

    diff =
      (realToFrac p - 1) * signum (n - z)
  in
    if sz == 99 then
      z + n
    else
      z + diff
```

That way it's using floating point for less of the calculation, and it will work for `Rational` too, even if it's not ideal.

chris-martin: I'm still doubting whether it's desirable to have `Rational` exponential ranges, because "shrinking" sort of loses its human relevance if the numerators and denominators are always huge numbers regardless of the size parameter.

```
> scaleExponentialFrac 80 (0 :: Rational) (100 :: Rational)
5721500312938077 % 140737488355328

> scaleExponentialFrac 2 (0 :: Rational) (100 :: Rational)
440089068475887 % 4503599627370496
```

jystic:
:shipit:  I think this is better than what we have, so I'm going to rebase and merge it, we can always improve the algorithm in the future. No sense bikeshedding over `Rational` imo.

jystic:
Rebase and merged https://github.com/hedgehogqa/haskell-hedgehog/commit/fca1700ce9bdd15002e21b44a6980f6183339564

PR #44, opened 2017-04-11
Add doctest to the build

jystic:


hedgehog/src/Hedgehog.hs
```diff
@@ -37,9 +37,9 @@
 --
 -- You can then load the module in GHCi, and run it:
 --
--- >>> tests
--- ━━━ Test.Example ━━━
---   ✓ prop_reverse passed 100 tests.
+-- > λ tests
```
moodmosaic: 👍 for the fancy λάμδα

jystic: Add it your `~/.ghc/ghci.conf` for extra points 🚀 
```
:set prompt "\ESC[34m\STX%s\n\ESC[0;33m\STXλ \ESC[m\STX"
```
screen shot 2017-04-12 at 1 23 38 pm

moodmosaic: It's there already + on my keyboard as well 😉 

c3-ufqzxuaaymmi

jystic: Nice 👍 fancy

PR #45, opened 2017-04-12
Remove redundant do-notation

moodmosaic:

PR #47, opened 2017-04-12
Remove redundant $

moodmosaic:
plus two redundant `do`s that slipped through #45.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -836,9 +836,9 @@ renderDoc color doc = do
     WL.indent 2 doc
 
 renderProgress :: MonadIO m => Maybe PropertyName -> Report Progress -> m String
-renderProgress name x = do
+renderProgress name x =
   renderDoc DetectColor =<< ppProgress name x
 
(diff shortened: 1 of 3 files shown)


hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -601,7 +601,7 @@ ppFailureReport name (FailureReport size seed _ inputs0 mlocation0 msg mdiff msg
         WL.vsep . WL.punctuate WL.line
     , with decls $
         WL.vsep . WL.punctuate WL.line . fmap ppDeclaration
-    , with msgs $
+    , with msgs
```
jystic: Hmm, I might be crazy, but despite being redundant I kind of prefer keeping this one because it makes all the `with` calls consistent

moodmosaic: Yeah, I thought the same on that one but wasn't really sure. I will put it back 👍

moodmosaic: This one is back now.

jystic:
👍 Rebased and merged

Issue #32, opened 2017-04-09
Support for Windows


moodmosaic:
I've converted a couple of QuickCheck properties to Hedgehog and they seem to be running fine on macOS:

![image](https://cloud.githubusercontent.com/assets/287532/24839079/e89882c6-1d5c-11e7-9217-ea051e246680.png)

However, when I try to run the same tests on Windows I'm getting:

```
*** Exception: <stdout>: hPutChar: invalid argument (invalid character)
```

---

In case it helps to troubleshoot, I used

* [this](https://github.com/hedgehogqa/haskell-hedgehog/blob/cc3af0e1ed573f0c6c484b0b979535bbc068e336/hedgehog/src/Hedgehog.hs#L25-L27) Template Haskell splice to run the properties in both macOS and Windows
* GHC 8.0.2 on macOS and GHC 7.10.3 on Windows

jystic:
I suspect it's the unicode symbols? The colours and in-place updates should get disabled because they're checking [hSupportsANSI](https://hackage.haskell.org/package/ansi-terminal-0.6.2.3/docs/System-Console-ANSI.html#v:hSupportsANSI)

hdgarrood:
I think the fix for this is usually along the lines of running `hSetEncoding stderr utf8` before attempting to print to the console. At least I remember a similar bug being resolved in the purescript compiler that way.

jystic:
> I think the fix for this is usually along the lines of running `hSetEncoding stderr utf8` before attempting to print to the console.

Nice, I'll give that a shot

moodmosaic:
I *may* be able to give this a shot as well, over the next few weeks.

thumphries:
I *think* we can treat this as a dupe of #110 - the discussion there is a little more thorough, so let's keep that one open and close this. Reopen or file a more specific issue if more Windows problems emerge.

PR #48, opened 2017-04-12
Keep the boss happy

jystic:
This is partially for amusement value 😆 (but in all seriousness, it's also so I can keep the default as displaying with colour)

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveFoldable #-}
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE DeriveTraversable #-}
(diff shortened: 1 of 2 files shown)


moodmosaic:
👍 I'll try this on Windows and see if it solves #32.

jystic:
I don't think it will, I suspect #32 is because of the unicode [ticks](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Internal/Report.hs#L679) and [crosses](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Internal/Report.hs#L658)

Will probably take a look at that one on the weekend

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -669,8 +682,17 @@ ppResult name (Report tests discards result) =
         ppTestCount tests <>
         "."
 
-useColor :: MonadIO m => m Bool
-useColor =
+detectMark :: MonadIO m => m Bool
```
thumphries: 💯

PR #50, opened 2017-04-12
Expose GroupName and PropertyName

chris-martin:
This adds `GroupName` and `PropertyName` to the `Hedgehog` module; these types are important to have when you're constructing groups explicitly instead of using TH.

I also added `Semigroup` instances because I found myself wanting to append text onto property names.


jystic:
👍

PR #60, opened 2017-04-26
Add helper for observing pure values

thumphries:
We already have `info` as something corresponding to QuickCheck's `counterexample`, but generated values get a really nice presentation out of the box. I'd like to be able to render pure values this way too, since subexpressions can be just as important as the generated value when trying to understand a counterexample.

In Jack and Hedgehog thus far I've used `forAll (pure x)` to achieve this. It's been really useful, so I'd like to recommend it as part of the API.

Not sure about the name. I was thinking something like `observe` would be nice?

@jystic @charleso

### DIFF
--- a/hedgehog/src/Hedgehog.hs
+++ b/hedgehog/src/Hedgehog.hs
@@ -72,6 +72,7 @@ module Hedgehog (
   -- * Test
   , forAll
   , info
+  , input
   , success
   , discard
   , failure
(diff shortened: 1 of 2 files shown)


thumphries:
Judging by the build failure, `input` is far too common a noun for an API like this :1st_place_medal:

charleso:
`log`?

Seems reasonable to me though.

jystic:
I'm thinking of renaming `info` to `footnote` and then calling this thing `annotate`, because it's annotating the source code.

jystic:
I've implemented this in https://github.com/hedgehogqa/haskell-hedgehog/pull/69

thumphries:
Nice! Thanks.

PR #61, opened 2017-04-26
Rename checkConcurrent -> checkParallel

jystic:
I think it's technically more correct to say that we're running the tests in parallel, as the goal is improved throughput, and there is no non-deterministic control flow.

I thought I would rename this before release as there is already a breaking change in this area due to the introduction of `$$(discover)`

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -350,6 +350,12 @@ checkSequential =
 
 -- | Check a group of properties in parallel.
 --
+--   /Warning: although this check function runs tests faster than/
+--   /'checkSequential', it should be noted that it may cause problems with/
+--   /properties that are not self-contained. For example, if you have a group/
+--   /of tests which all use the same database table, you may find that they/
+--   /interfere with each other when being run in parallel./
+--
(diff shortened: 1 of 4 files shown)


thumphries:
I think `parallel` has safety connotations that `concurrent` does not. Tests can still interfere with one another. Could we add a written warning to the Haddock alongside this change?

jystic:
I'm not 100% on this change, happy to keep the original name if it's clearer

thumphries:
We need a warning either way IMO, and the name change makes sense as long as the user thinks carefully before using the parallel facility.

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -350,6 +350,12 @@ checkSequential =
 
 -- | Check a group of properties in parallel.
 --
+--   /Warning: although this check function runs tests faster than/
+--   /'checkSequential', it should be noted that it may cause problems with/
+--   /properties that are not self-contained. For example, if you have a group/
+--   /of tests which all use the same database table, you may find that they/
+--   /interfere with each other when being run in parallel./
```
thumphries: Seems clear enough to me.

PR #63, opened 2017-04-27
Remove redundant do-notation

moodmosaic:


jystic:
I don't think it makes sense to proceed with these "remove redundant do-notation" PRs as it really isn't a big deal whether it's there or not. If we are serious about not having redundant do-notation I think we need a build time tool, otherwise it's just going to keep happening.

moodmosaic:
Yeah... I was actually thinking something similar. Not sure if it's worth the effort though(?) These 'suggestions' pop up form HLint every time I pull the latest changes.

PR #65, opened 2017-05-02
Remove use of Typeable / allow custom render function

jystic:
It was only for 7.10 anyway, and it turns out that some type level hackery can't be used with a `Typeable` constraint. This, in conjunction with a custom render function, will allow those situations to work properly.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -35,6 +35,7 @@ module Hedgehog.Internal.Property (
   , Failure(..)
   , Diff(..)
   , forAll
+  , forAllWith
   , info
   , discard
   , failure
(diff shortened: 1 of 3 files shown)


charleso:
👍

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -331,10 +331,20 @@ writeLog =
...
+  withFrozenCallStack $ forAllWith showPretty gen
+
+-- | Generates a random input for the test by running the provided generator.
+--
+--   /This is a the same as 'forAll' but allows the user to provide a custom/
+--   /rendering function. This is useful for values which don't have a/
+--   /'Show' instance./
+--
+forAllWith :: (Monad m, HasCallStack) => (a -> String) -> Gen m a -> Test m a
+forAllWith render gen = do
```
moodmosaic: > This is a the same as 'forAll' but allows the user to provide a custom/
> /rendering function. This is useful for values which don't have a/
> /'Show' instance.

Perhaps, *rendering* or *show* could be part of the function name? Perhaps, `forAllWith` is too general? 😕 Hard for me to come up with a good alternative though.

jystic: Yeah I was wondering about that, I have called this `forAllRender` in the past, but I was thinking that potentially this may grow options other than just the rendering function in the future. I have called all the functions with take extra options `*With` so far in hedgehog, but I'm not tied to this necessarily.

jystic:
Happy to rename this later if we think of something better

PR #73, opened 2017-05-06
Add ascii/latin1/unicode character generators

jystic:
This adds some generators for various character sets, and a property test which fails with the current version of `text` on Hackage as an example usage.

The test was inspired by https://github.com/bos/text/issues/176 and the generator choices are similar to the ones which have recently been added to QuickCheck using `Arbitrary` instances for `Char` newtypes.

screen shot 2017-05-06 at 5 42 32 pm

### DIFF
--- a/hedgehog-example/test/Test/Example/Basic.hs
+++ b/hedgehog-example/test/Test/Example/Basic.hs
@@ -6,6 +6,9 @@ module Test.Example.Basic where
 
 import           Control.Monad (guard)
 
+import qualified Data.List as List
+import qualified Data.Text as Text
+
 import           Hedgehog
 import qualified Hedgehog.Gen as Gen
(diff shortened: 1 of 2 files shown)


hedgehog-example/test/Test/Example/Basic.hs
```diff
@@ -251,6 +254,24 @@ prop_record =
     x === y
 
 ------------------------------------------------------------------------
+-- Example 6 - Text.takeEnd
+
+prop_takeEnd :: Property
+prop_takeEnd =
+  property $ do
+    xs <- forAll $ Gen.string (Range.linear 0 100) Gen.unicode
```
jystic: If this is changed from `Gen.unicode` to `Gen.latin1` then the test will start passing

PR #76, opened 2017-05-10
Use (DFFF)16/(57343)10 as the low surrogate

moodmosaic:
I came across this while porting https://github.com/hedgehogqa/haskell-hedgehog/pull/73/ to the F# version: https://github.com/hedgehogqa/fsharp-hedgehog/pull/96/.

See also:
- https://en.wikipedia.org/wiki/Universal_Character_Set_characters#Surrogates
- https://referencesource.microsoft.com/#mscorlib/system/char.cs,913


jystic:
👍 nice catch

PR #77, opened 2017-05-11
FIxes for upcoming ghc version

erikd:


hedgehog/src/Hedgehog/Internal/TH.hs
```diff
@@ -13,6 +14,9 @@ import           Hedgehog.Internal.Discovery
 import           Hedgehog.Internal.Property
 
 import           Language.Haskell.TH
+#if MIN_VERSION_template_haskell(2, 12, 0)
+                    hiding (TExpQ)
+#endif
 import           Language.Haskell.TH.Syntax
```
jystic: This was lazy on my part, they should have been explicit imports from the beginning (then `CPP` wouldn't be required here.)

erikd: I can make all the import qualified or explicit, either just for this file or for all files. Your choice.

hedgehog/src/Hedgehog/Internal/TH.hs
```diff
@@ -12,8 +12,8 @@ import qualified Data.Ord as Ord
 import           Hedgehog.Internal.Discovery
 import           Hedgehog.Internal.Property
 
-import           Language.Haskell.TH
-import           Language.Haskell.TH.Syntax
+import           Language.Haskell.TH (Exp (..), Q, TExp, location, runIO)
+import           Language.Haskell.TH.Syntax (Loc (..), mkName, unTypeQ, unsafeTExpCoerce)
```
jystic: Just a minor nit, can you remove the space between `Exp` and `(..)` to match the style of the rest of the project (same for `Loc`)

jystic:
👍 after nitpick

hedgehog/src/Hedgehog/Internal/TH.hs
erikd: Done!

erikd:
Fixed nitpicks, but I can't merge this.

jystic:
Rebased and merged 🌵

PR #78, opened 2017-05-15
Add Test.Hedgehog.Text to other modules

amarpotghan:
So that it's included in distribution. Currently build is failing on hydra: http://hydra.nixos.org/build/51844414/nixlog/1

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -117,7 +117,7 @@ test-suite test
     test
 
   other-modules:
-      Test.Hedgehog.Text
+    Test.Hedgehog.Text
 
   build-depends:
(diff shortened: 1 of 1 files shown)


hedgehog/hedgehog.cabal
```diff
@@ -116,6 +116,9 @@ test-suite test
   hs-source-dirs:
     test
 
+  other-modules:
+      Test.Hedgehog.Text
```
jystic: Happy to add this, would you be able to align it at the same level as the modules above (so, two space indent):
```
  other-modules:
    Test.Hedgehog.Text
```

jystic:
Rebase and merged 👍

PR #80, opened 2017-05-16
Increase upper bounds of concurrent-output and wl-pprint-annotated (Fixes #79)

### LINKED ISSUE #79: Lift the upper bounds of concurrent-output and wl-pprint-annotated
I would send a PR if you agree with the following:

```
wl-pprint-annotated  >= 0.0 && < 0.2
concurrent-output   >= 1.7  && < 2.0
```

That way both libraries will be upgraded to the latest versions. What do you think?

amarpotghan:


jystic:
Rebased and merged 👍

PR #83, opened 2017-06-10
Distributive class for pulling a transformer out to the top level

jystic:
Perhaps adding a type class here is overkill, but I think we need the type family instances to make it manageable to use, so :man_shrugging: 

I think I'm going to go for it and see what happens.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -48,6 +48,7 @@ module Hedgehog.Internal.Property (
 
   , liftEither
   , liftExceptT
+  , withExceptT
   , withResourceT
 
   -- * Internal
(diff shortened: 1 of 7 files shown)


hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -449,6 +469,13 @@ liftExceptT :: (Monad m, Show x, HasCallStack) => ExceptT x m a -> Test m a
 liftExceptT m =
   withFrozenCallStack liftEither =<< lift (runExceptT m)
 
+-- | Fails the test if the 'ExceptT' is 'Left', otherwise returns the value in
+--   the 'Right'.
+--
+withExceptT :: (Monad m, Show x, HasCallStack) => Test (ExceptT x m) a -> Test m a
+withExceptT m =
+  withFrozenCallStack liftEither =<< runExceptT (distribute m)
```
jystic: I'm aware that this clashes with `Control.Monad.Trans.Except.withExceptT` but I'm not sure what else to call it. Whatever it's called it should be the same prefix as `withResourceT` I think.

jystic: I'm going to go with this for now, if we decide to pick a different name it will be a breaking change anyway because of `withResourceT` so I don't see a reason to hold this up.

PR #85, opened 2017-06-11
Add roundtrip test from Tim's blog post

jystic:
I hope @thumphries doesn't mind, I turned his [blog post](http://teh.id.au/posts/2017/06/07/round-trip-property/) in to an example.

While I was doing so I noticed that the `tripping` function could be improved a bit, I added an output with the intermediate value:
screen shot 2017-06-11 at 11 37 54 am

I understand that the intermediate value may not always be something supporting `Show` and could potentially be large, but I think this behaviour is a nice default. I think longer term we should have a `trippingWith` function which is a bit more configurable, but this is ok for now.


thumphries:
:+1: glad to share it!

PR #86, opened 2017-06-11
Release 0.3

jystic:


jystic:
Not sure why the @username and #ticket things don't work in the markdown, oh well

moodmosaic:
>Not sure why the @username and #ticket things don't work in the markdown, oh well

Maybe they just need to be regular [markdown links](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links)? Happy to send a pull request if you want 😃

jystic:
Yeah it's bizarre, they work in github comments and on realease pages, but it seems not in documents. I worry that actual links would mess up the readability of the markdown. I guess we can give it a try. Maybe having the link text at the bottom of the page would work (see main readme source).

moodmosaic:
Sure, I was thinking more like having them as reference-style links, as in the main readme source.

Like so:

```
[I'm a reference-style link][Arbitrary case-insensitive reference text]

[arbitrary case-insensitive reference text]: https://hedgehog.qa/
```

Will do a pull request and see what happens 😃

jystic:
> Sure, I was thinking more like having them as reference-style links, as in the main readme source.

Ah yep, that's what I meant 👍

moodmosaic:
I've opened https://github.com/hedgehogqa/haskell-hedgehog/pull/93 😃

PR #89, opened 2017-06-25
Abstract state machine testing

jystic:
A colleague pointed me to the release of the very cool [quickcheck-state-machine](https://github.com/advancedtelematic/quickcheck-state-machine) so I decided I'd better stop sitting on this and actually publish it!

It's still a little raw in the sense that you need to write a few boilerplate instances for your commands, and there isn't any parallel testing yet. I think [parallel](http://www.cse.chalmers.se/~nicsma/papers/finding-race-conditions.pdf) testing should fit in to the design without any issues however.

I rejigged the `Hedgehog.Gen` and `Hedgehog.Range` modules so they're just re-exporting from internals in order to add `Gen.actions` as a thing. (sorry @charleso :# ! it's all about syntax)

Here's some sample output from the two examples I've knocked together so far.

### Process Registry

screen shot 2017-06-25 at 9 37 27 pm

### Mutable References

screen shot 2017-06-25 at 9 37 50 pm

/cc @moodmosaic @thumphries @charleso


hedgehog/src/Hedgehog/Internal/State.hs
```diff
@@ -0,0 +1,474 @@
...
+    -- | Executes a command using the arguments generated by 'commandGen'.
+    --
+    , commandExecute ::
+        input Concrete -> Test m output
+
+    -- | A set of callbacks which provide optional command configuration such
+    --   as pre-condtions, post-conditions and state updates.
+    --
+    , commandCallbacks ::
+        [Callback input output m state]
```
charleso: I have mixed feelings about this. Alternatively would it be possible to provide extra constructor functions with defaults? Would that be better/worse?

jystic: It's not really possible to do the defaults thing because of the existentials.

I convinced myself that this is the least bad solution. I liked it because it has the benefit of being able to abstract over the functions that implement the model itself. For example if you have pre- & post-conditions which are shared between a number of commands then you can pull them out and just append it to your list.

Is it just the name which gives a bad feel? I took the callback terminology from the Erlang QuickCheck, but happy to change to something else.

charleso: As discussed, was just curious. As long as it's not just for convenience but because it's a better API, which is sounds like it is. Just ignore me. :)

jystic: I guess we'll find out! really hard to say without using it in anger, we can always change it

PR #90, opened 2017-06-27
Release 0.4

jystic:
The state machine stuff is still a bit experimental, but we'll never get it solid if we don't use it, time for a release! 🔥


charleso:
🎂

moodmosaic:
Good idea! 💯

jystic:
:shipit: [0.4](http://hackage.haskell.org/package/hedgehog-0.4)

moodmosaic:
It'll be now slightly faster/easier to test #89 🎉

PR #91, opened 2017-06-28
Types are a thing

jystic:
Found this while using the new state machine stuff, if you have two different return types then you can end up with a type error while shrinking. This fixes that issue.


charleso:
👍

moodmosaic:
👍

PR #94, opened 2017-06-30
Easier to use variables for state machine testing

jystic:
This adds an extra data type for working with state machine variables. It frees the user from having to derive `Eq`, `Ord`, `Show` by hand for data structures which contain variables.

This PR also adds an extra argument to the `Ensure` callback which is the previous state. Having the previous and the current state can sometimes be useful when writing post-conditions.


charleso:
💯

PR #96, opened 2017-07-02
Separate testing and input generation concepts

jystic:
I got stuck implementing parallel state machine testing as `Test` was not `MonadBaseControl` so I couldn't fork it to a new thread. This change makes a separation between things that support testing (`Test`) and things that support testing and can generate input (`TestGen`). This means that we can fork a thread in `Test` even though we can't do so for `TestGen`. It also clears up the strange  `ResourceT` behaviour somewhat.

I'm not really happy with the names `Test` and `TestGen`, would probably prefer `Test` and `Property`, but then I don't know what to call the top level testable thing.

I'm also tossing up on the names of the `eval` functions, I named them this way so that `evalExceptT` looks like `evalStateT` from `transformers`. But really `eval` itself is more similar to `evaluate` in `base:Control.Exception`

! @charleso, I wouldn't mind a discussion about this some time


charleso:
👍

PR #98, opened 2017-07-15
Parallel state machine testing

jystic:
This adds supports for parallel state machine testing. The shrinking could be vastly improved but it's at least a start.

I have rejigged some of the monad transformers because it now becomes useful to have a pure test which can be used in the post-condition for state machines.

screen shot 2017-07-15 at 2 04 25 pm

### DIFF
--- a/hedgehog-example/test/Test/Example/References.hs
+++ b/hedgehog-example/test/Test/Example/References.hs
@@ -136,8 +136,8 @@ instance HTraversable IncRef where
   htraverse f (IncRef ref) =
     IncRef <$> htraverse f ref
 
-incRef :: (Monad n, MonadIO m) => Command n m State
-incRef =
+incRef :: (Monad n, MonadIO m) => Int -> Command n m State
+incRef n =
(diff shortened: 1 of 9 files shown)


charleso:
💯

jystic:
Possibly a better example, it's not a minimal shrink, but it does pretty clearly find that `IncRef` is not atomic:

screen shot 2017-07-15 at 2 27 19 pm

hedgehog-example/test/Test/Example/References.hs
```diff
@@ -166,14 +166,33 @@ incRef =
...
+        , incRef 2
+        ]
 
     executeSequential initialState actions
 
+prop_references_parallel :: Property
+prop_references_parallel =
+  withTests 1000 . withRetries 10 . property $ do
+    actions <- forAll $
+      Gen.parallel (Range.linear 1 50) (Range.linear 1 10) initialState [
```
moodmosaic: 👍

PR #99, opened 2017-07-16
MonadGen class to make working with transformers easier

jystic:
This renames `Gen` to `GenT` and introduces the `MonadGen` class.

The `MonadGen` class has instances for a number of transformers, including `ReaderT` and `StateT`. This means that it's now possible to use these transformers on the outside of a `Gen`, as such I think it will be much more common to have a pure `Gen`, so I've renamed the transformer variant to `GenT` and changed `forAll` to expect a pure `Gen`. Folks can use `forAllT` if they need effects in their generators.

I'm super happy with how this worked out, I think writing generators will be a lot cleaner now.


hedgehog-example/test/Test/Example/STLC.hs
```diff
@@ -166,11 +165,11 @@ genType =
...
 
-genWellTypedExpr :: Monad m => Type -> Gen m Expr
-genWellTypedExpr want =
-  hoist (generalize . flip runReaderT mempty) (genWellTypedExpr' want)
+genWellTypedExpr :: Type -> Gen Expr
+genWellTypedExpr =
+  flip runReaderT mempty . genWellTypedExpr'
 
-genWellTypedExpr' :: Type -> Gen (Reader (Map Type [Expr])) Expr
+genWellTypedExpr' :: Type -> ReaderT (Map Type [Expr]) Gen Expr
```
jystic: /cc @thumphries

hedgehog-example/test/Test/Example/STLC.hs
```diff
@@ -166,11 +165,11 @@ genType =
 
 ------------------------------------------------------------------------
 
-genWellTypedExpr :: Monad m => Type -> Gen m Expr
-genWellTypedExpr want =
-  hoist (generalize . flip runReaderT mempty) (genWellTypedExpr' want)
+genWellTypedExpr :: Type -> Gen Expr
+genWellTypedExpr =
+  flip runReaderT mempty . genWellTypedExpr'
```
jystic: /cc @thumphries

hedgehog-example/test/Test/Example/STLC.hs
thumphries: 👍  more intuitive

hedgehog-example/test/Test/Example/STLC.hs
thumphries: 💯 💯

PR #100, opened 2017-07-16
Gen.sample now only gives a single result back

jystic:
This changes `Gen.sample` to return a single result instead of a list. Users can always use `replicateM 10 Gen.sample` if they want the old behaviour. /cc @erikd

I also changed the sampling functions to work over `Gen` instead of `GenT` because otherwise users would need to say `Gen.sample (Gen.lift gen)` if they wanted to sample vanilla generators, which I think is going to be the common case.


erikd:
👍 on 🍏

moodmosaic:
👍 👍

PR #102, opened 2017-07-16
Use haddock not-home option for internals

thumphries:
At the moment, clicking around on Haddock references to the public API leads you into the internal modules.

The `not-home` option makes Haddock link to the public API for publicly exported identifiers. It still links to the internals for things that aren't otherwise exported.


thumphries:
It took me 1000 hours to make OS X sed insert a line at the start of a file.

jystic:
💯 💯 💯 

I never knew about this, it was annoying me too

moodmosaic:
Neat! 🥇

PR #103, opened 2017-07-16
Support for using QuickCheck generators in Hedgehog and vice versa

jystic:
This allows the use of QuickCheck generators and Arbitrary instances in Hedgehog, and also the use of Hedgehog generators inside QuickCheck.

Unfortunately I can't think of a way to get shrinking when using Hedgehog generators inside QuickCheck, but the other way is possible.


moodmosaic:
💯 This is awesome, lookibg forward trying this out on a couple of QuickCheck codebases.

PR #112, opened 2017-07-26
Added ability to use comparison functions other than ==

Kheldar:


jystic:
Thanks for the PR, will get a chance to look at this properly on the weekend.

Kheldar:
Have you had a chance to look at this?

jystic:
Sorry, not yet, I've been too busy, will get to it as soon as I can

Kheldar:
Any chance this could be looked at?

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -545,13 +557,36 @@ infix 4 ===
...
-(===) x y = do
-  ok <- withFrozenCallStack $ eval (x == y)
+(===) = compareWith (==)
+
+infix 4 /==
+
+-- | Fails the test if the two arguments provided are equal.
+--
+(/==) :: (MonadTest m, Eq a, Show a, HasCallStack) => a -> a -> m ()
+(/==) = compareWith (/=)
```
puffnfresh: This would do a diff when values are equal. Not ideal.

jystic:
Diffs only make sense for equality like comparisons.

jystic:
Similar functionality merged in https://github.com/hedgehogqa/haskell-hedgehog/pull/196

PR #115, opened 2017-08-17
Remove unnecessary usages of 'do' keyword

gwils:


moodmosaic:
👍

---

FWIW, we've done similar cleanups in the past. The last one is [here](https://github.com/hedgehogqa/haskell-hedgehog/pull/63), with a [comment](https://github.com/hedgehogqa/haskell-hedgehog/pull/63#issuecomment-297877400) by @jystic:
>[..] If we are serious about not having redundant do-notation I think we need a build time tool, otherwise it's just going to keep happening.

Maybe we should work on that instead?

gwils:
I have hlint installed on my system, and I just invoke it with `hlint .` at a project's root directory.
hlint currently gives 90 hints, or 86 if this pull request is merged.

In this case, I just noticed an unnecessary do while browsing the source and thought I would weed them out :)

ocharles:
> Maybe we should work on that instead?

Maybe, but remember that perfect is the enemy of good. Even if it's just an occasional drive by clean up PR, it's better than doing nothing.

gwils:
Wondering where we go from here? Travis seems like it had a bad time on GHC 8.2.1. I'm not sure why.

thumphries:
This PR is probably not going to be merged (based on the outcome last time), let's leave this sort of thing unless it's enforced project-wide. Sorry for letting it hang open for so long.

PR #116, opened 2017-08-22
Bump ansi-terminal bound

TomMD:


jystic:
Thanks for the heads up, I've just edited the bounds on Hackage so 0.5 should be installable with these constraints. Will merge this so that we don't regress.

PR #124, opened 2017-10-25
Correct mixGamma oddness check.

markhibberd:
See mixGamma http://gee.cs.oswego.edu/dl/papers/oopsla14.pdf on page 13.

Also see jane streets implementation for comparison https://github.com/janestreet/core_kernel/blob/master/src/splittable_random.ml#L87.

@jystic On a side note, there is a dedicated place for people who write bit masks in decimal, I dearly hope you have someone else to blame.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Seed.hs
+++ b/hedgehog/src/Hedgehog/Internal/Seed.hs
@@ -192,7 +192,7 @@ mixGamma x =
     y = mix64variant13 x .|. 1
     n = popCount $ y `xor` (y `shiftR` 1)
   in
-    if n < 24 then
+    if n >= 24 then
       y `xor` (-6148914691236517206)
     else
       y


moodmosaic:
I see the same check in the [F# version](https://github.com/hedgehogqa/fsharp-hedgehog/blob/b009cfb9e2317d2fb4d7b487feb57406e1872867/src/Hedgehog/Seed.fs#L90) so that must be my fault while porting the code from the paper. I’m not 100% sure about the bit masks in the Haskell version, but happy to take the blame there as well 👍

markhibberd:
@moodmosaic More jest then blame, so apologies, I was just wanting to give @jystic a hard time. Bugs happen.

moodmosaic:
@markhibberd but thanks for catching that one! I've no idea how it slipped through :cowboy_hat_face:

jystic:
> @jystic On a side note, there is a dedicated place for people who write bit masks in decimal, I dearly hope you have someone else to blame.

I've been racking my brain to figure out why I did this, as it's clearly not in the F# version. I was playing in GHCi and I think this is probably why I did it:
```
λ 0xaaaaaaaaaaaaaaaa :: Int64
<interactive>:13:1: warning: [-Woverflowed-literals]
    Literal 12297829382473034410 is out of the Int64 range -9223372036854775808..9223372036854775807
```

I think it might make more sense to just disable that warning for this file though.

jystic:
👍

jystic:
👍

moodmosaic:
FWIW, the Java version has the opposite (the one we had originally):
https://github.com/JetBrains/jdk8u_jdk/blob/1c45369790de2b97ca2203db4b5c41922c54d9b9/src/share/classes/java/util/SplittableRandom.java#L212

PR #129, opened 2017-11-14
Add mapTestT operator.

tmcgilchrist:
cc @charleso @jystic 

Needed this operator for writing some amazonka IO tests, unfortunately the code is internal Ambiata right now but the general pattern is
```haskell
prop_standard_api_no_duplicates = property $ do
  test . mapTestT runAWSDefaultRegion $ do
    env <- lift testAWSEnv
    # Run amazonka things
    # assert using (===)

```

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -451,9 +451,9 @@ runTest :: Test a -> (Either Failure a, [Log])
 runTest =
   runIdentity . runTestT
 
-mapTestT :: (forall b. m b -> n b) -> TestT m a -> TestT n a
+mapTestT :: Monad m => (b -> m a) -> TestT m b -> TestT m a
 mapTestT f =
-  mkTestT . f . runTestT
+  mkTestT . (=<<) (\(e, l) -> fmap (flip (,) l) $ traverse f e) . runTestT
 
 -- | Log some information which might be relevant to a potential test failure.
 --
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -75,6 +75,7 @@ module Hedgehog.Internal.Property (
   , mkTestT
   , runTest
   , runTestT
+  , mapTestT
   ) where
 
 import           Control.Applicative (Alternative(..))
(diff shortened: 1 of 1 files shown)


hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -450,6 +451,10 @@ runTest :: Test a -> (Either Failure a, [Log])
 runTest =
   runIdentity . runTestT
 
+mapTestT :: (forall b. m b -> n b) -> TestT m a -> TestT n a
```
charleso: FWIW this might be slightly more powerful/useful if it was:

```hs
mapTestT :: Functor m => (m a -> n b) -> TestT m a -> TestT n b
mapTestT f =
  mkTestT . first (fmap f) . runTestT
```

charleso: Sorry, maybe something more like:

```hs
mapTestT :: Applicative m => (m a -> n b) -> TestT m a -> TestT n b
mapTestT f =
  mkTestT . (=<<) (\(e, l) -> fmap (flip (,) l) $ traverse f e) . runTestT
```

charleso:
👍 

I defer to Jacob all things hedgehog though

hedgehog/src/Hedgehog/Internal/Property.hs
tmcgilchrist: The more general version seems better.

nhibberd: Yeah without the more general version it's just the same as `hoist`

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -450,6 +451,10 @@ runTest :: Test a -> (Either Failure a, [Log])
 runTest =
   runIdentity . runTestT
 
+mapTestT :: Monad m => (b -> m a) -> TestT m b -> TestT m a
+mapTestT f =
+  mkTestT . (=<<) (\(e, l) -> fmap (flip (,) l) $ traverse f e) . runTestT
```
jystic: I think this new one is actually not that useful, as the `m` can't be changed.

One alternative which gives more power than `hoist` is:
```hs
mapTestT :: (m (Either Failure a, [Log]) -> n (Either Failure b, [Log])) -> TestT m a -> TestT n b
mapTestT f =
  mkTestT . f . runTestT
```

Not sure how useful that one is?

jystic:
I should point out that the thing you were initially trying to do is already supported via `hoist`:
```hs
prop_standard_api_no_duplicates = property $ do
  test . hoist runAWSDefaultRegion $ do
    env <- lift testAWSEnv
    # Run amazonka things
    # assert using (===)
```

tmcgilchrist:
So it does, not much point in this function then.

PR #130, opened 2017-11-21
only invoke setNumCapabilities when using the -threaded runtime

ekmett:
This clears up some warning text.


jystic:
Thanks, will merge when I’m back in the office :+1:

PR #131, opened 2017-11-28
Simplify dependency on semigroups.

gwils:
There's no need to exclude the package on older GHC versions. The package itself
does this with CPP.


jystic:
Doesn't this mean that there is now an unnecessary dependency on semigroups in the case of ghc >= 8.0?

gwils:
`semigroups` turns itself into basically an empty package on `ghc >= 8.0`.
I think depending on an empty package is preferable to complicating the build by switching on GHC versions. I see that as a duplication of work that `semigroups` has already done.

It's a matter of preference. You might feel differently :)

thumphries:
Let's keep this as is.

PR #139, opened 2017-12-22
Added a /== operator to express that 2 values must be different

etorreborre:
Hi @thumphries, @jystic, here is a proposal for a `/==` operator. 

Hmm I can see that stylish-haskell has made the diff larger than it should. Do you have any specific formatting rules?

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -55,6 +55,7 @@ module Hedgehog.Internal.Property (
   , success
   , assert
   , (===)
+  , (/==)
 
   , eval
   , evalM
@@ -551,6 +552,22 @@ infix 4 ===
     success
   else
     withFrozenCallStack $ failDiff x y
+
+infix 4 /==
+
+-- | Fails the test if the two arguments provided are equal.
+--
+(/==) :: (MonadTest m, Eq a, Show a, HasCallStack) => a -> a -> m ()
+(/==) x y = do
+  ok <- withFrozenCallStack $ eval (x /= y)
+  if ok then
+    success
+  else
+    withFrozenCallStack $
+      failWith Nothing $ unlines [
+          "━━━ Both equal to ━━━"
+        , showPretty x
+        ]
 
 -- | Fails the test if the value throws an exception when evaluated to weak
 --   head normal form (WHNF).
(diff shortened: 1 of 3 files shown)


thumphries:
Thanks for the PR, Eric!

It'd be nice to get that stylish-haskell noise out of the PR. Small diffs are good. Usually we try to avoid context-dependent whitespace to keep the diffs small. Besides that, just try to mimic the rest of the repo's style.

etorreborre:
I assumed so, here is the trimmed PR.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -552,6 +553,22 @@ infix 4 ===
...
+
+-- | Fails the test if the two arguments provided are equal.
+--
+(/==) :: (MonadTest m, Eq a, Show a, HasCallStack) => a -> a -> m ()
+(/==) x y = do
+  ok <- withFrozenCallStack $ eval (x /= y)
+  if ok then
+    success
+  else
+     withFrozenCallStack $
```
thumphries: Could this be reformatted like https://github.com/etorreborre/haskell-hedgehog/blob/c6b44787554e9703e592e4f5a787b85da41036bb/hedgehog/src/Hedgehog/Internal/Property.hs#L501 ?

i.e. move `withFrozenCallStack` back slightly, reindent the `failWith` block

thumphries:
Sorry it took so long to fully consider this. Everyone's been busy.

An operator for this is a good idea. We weren't totally sure if it should be `(=/=)` or `(/==)`. Since nobody felt too strongly either way, I'll merge this when the formatting is fixed.

thumphries:
(`(/==)` seems self-consistent with `(===)` to me, but it depends on whether you see the extra `=` as being added to the end or to the front)

puffnfresh:
I'd like to see operators generalised a bit, too. Can we have less than, greater than, custom matchers with nice diffing?

etorreborre:
@thumphries I fixed the reformatting. I also don't feel strongly about `/==` vs `=/=`. Maybe `/==` is just easier to type since you don't alternate symbols (if one needs something to decide).

etorreborre:
@puffnfresh can we have this in a different PR, I don't have much time for OSS dev these days :-)?

puffnfresh:
@etorreborre yeah just saying, I'd like to see that.

PR #140, opened 2018-01-12
Depend on `semigroups`

LightAndLight:
Conditional dependency on `semigroups` causes some problems for tools that don't know now to deal with `if`s in cabal files. I can't build `hegdgehog` on the `ghc7102` or `ghc7103` package sets, because the tools don't know how to account for the conditional dependency. It's okay to just always depend on semigroups - see https://qfpl.io/posts/semigroups/

Please [squash this PR](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) when you merge it :)

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -128,9 +128,6 @@ test-suite test
     , base                            >= 3          && < 5
     , containers                      >= 0.4        && < 0.6
     , pretty-show                     >= 1.6        && < 1.7
+    , semigroups                      >= 0.16       && < 0.19
     , text                            >= 1.1        && < 1.3
     , transformers                    >= 0.3        && < 0.6
-
-  if impl(ghc < 8.0)
-    build-depends:
-      semigroups                      >= 0.16       && < 0.19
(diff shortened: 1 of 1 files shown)


LightAndLight:
I'm not sure why the tests have failed. I'll look into it later

gwils:
It looks like the jobs timed out.

thumphries:
That's interesting, which tool fails to understand Cabal files? `cabal2nix` or something? Does that cause problems more widely? Lots of packages use conditionals...

This seems harmless, we should probably merge it given the additional context.

(I don't have the perms, you will have to wait a little longer)

nhibberd:
> It's okay to just always depend on semigroups

This is still adding a redundant dependency for > 8

LightAndLight:
@thumphries Yeah, it's `cabal2nix`. It currently just ignores conditional dependencies.
@nhibberd Okay. Is one of the goals of this project to minimize the number of external dependencies? I use GHC8+, and the thought of an extra dependency doesn't bother me.

LightAndLight:
> In any case this is a Cassava kind of situation

Yeah I guess it is, although I wouldn't kick up such a fuss if you said no ;)

thumphries:
Sorry that took so long. Perms sorted now

PR #145, opened 2018-02-01
Add GHC 8.4.1 as an allowed failure to Travis

gwils:
We should prepare for the quickly approaching GHC 8.4 release. This adds the pre-release version to travis, but as an allowed failure for now.


thumphries:
👍  thanks george

PR #146, opened 2018-02-01
Remove outdated point releases of GHC from Travis

gwils:


nhibberd:
I don't see any harm in leaving these as current builds? Unless I am missing something

gwils:
Travis will only build so many at a time, so in the interest of prompt feedback on my branches and pull requests, I'd prefer to remove obsolete point versions.
It's not a huge deal either way.

thumphries:
Travis does 5 builds at a time. However we take between two days and three months to look at PRs, so it probably doesn't matter much

moodmosaic:
Yes, I think we should leave those in for now.

gwils:
Ok, I'll close this and add those versions to #149

PR #149, opened 2018-02-02
Use new build for travis

gwils:
This uses `cabal new-build` for Travis, as generated by hvr's script. I believe this will sidestep the travis 50 minute timeout problem we've been having in Travis.

Some things to note:
* I've trimmed the old point-versions as in my other PR. 7.10.2 in particular doesn't seem to work with hvr's script and the `cabal.project` model.
* hedgehog-example has been reshuffled a little to put most of the code in the library section. This keeps `cabal check` happy. I think cabal check is worthwhile to have on for hedgehog and hedgehog-quickcheck, so I think it's reasonable to pay the price of this little reshuffle in -example.

Some small modifications have been made from the output of hvr's script:
* `travis_wait` has been added to the call to cabal test, as the tests can run for >10 minutes without output on 7.10.3
* INSTALLED has been manually set to false. In these tests, this allows cabal to generate plans that disagree with the ghc boot libraries for the given version. transformers 0.4.x shipped as a boot library with 7.10, so we would need to support transformers 0.4 to fix this. I intend to do that soon, but I thought I'd put this PR up first.


thumphries:
Thanks for the PR!

> I believe this will sidestep the travis 50 minute timeout problem we've been having in Travis.

That sounds appealing - but does it? Do you have build timing numbers we can consider?

On a cold cache this saves us from recompiling mafia, but we could fix that in other ways.

gwils:
I seem to have successfully busted the cache: https://travis-ci.org/gwils/haskell-hedgehog/jobs/336375870#L482
and it seems to have run pretty quickly.

If you merge #150 (support for transformers 0.4) I'll be able to leave INSTALLED set to true in this script.
I'll resolve the conflict in a moment.

thumphries:
OK, this seems like a win overall. Let's get #150 in first.

hedgehog-example/hedgehog-example.cabal
```diff
@@ -22,28 +22,29 @@ cabal-version:
   >= 1.8
 build-type:
   Simple
+tested-with:
+  GHC == 7.10.3, GHC == 8.0.2, GHC == 8.2.2, GHC == 8.4.1
```
moodmosaic: Nitpick:

I think this should be formatted as

```
  GHC == 7.10.3
, GHC == 8.0.2
, GHC == 8.2.2
, GHC == 8.4.1
```

hedgehog-example/hedgehog-example.cabal
```diff
@@ -22,28 +22,29 @@ cabal-version:
...
+
+  exposed-modules:
+    Test.Example.Basic,
+    Test.Example.Exception,
+    Test.Example.QuickCheck,
+    Test.Example.References,
+    Test.Example.Registry,
+    Test.Example.Resource,
+    Test.Example.Roundtrip,
+    Test.Example.STLC
```
moodmosaic: Nitpick:

I think this should be formatted as

```
  Test.Example.Basic
, Test.Example.Exception
, Test.Example.QuickCheck
, Test.Example.References
, Test.Example.Registry
, Test.Example.Resource
, Test.Example.Roundtrip
, Test.Example.STLC
```

gwils:
@moodmosaic I have addressed your nitpicks. Thanks for pointing them out :)
I've also added more GHC versions here since we didn't end up going with #146

moodmosaic:
@gwils :+1: :guitar: :rocket:

thumphries:
Thanks George. Hopefully it works out better than it did last time 😅

PR #150, opened 2018-02-02
Support transformers 0.4

gwils:
Here's some CPP to support transformers 0.4. This should help some users using GHC 7.10 since transformers 0.4 is a boot library on that version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ install:
   - BENCH=${BENCH---enable-benchmarks}
   - TEST=${TEST---enable-tests}
   - HADDOCK=${HADDOCK-true}
-  - INSTALLED=${INSTALLED-false}
+  - INSTALLED=${INSTALLED-true}
   - GHCHEAD=${GHCHEAD-false}
   - travis_retry cabal update -v
   - "sed -i.bak 's/^jobs:/-- jobs:/' ${HOME}/.cabal/config"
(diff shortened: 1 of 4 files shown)


gwils:
Supporting boot libraries for a given GHC version is A Good Thing because any user who depends on ghc-the-library will be forced to use those versions. So at the moment nobody on GHC 7.10.3 who depends on the ghc library will be able to use hedgehog, which is a shame.

I think this CPP is a small price to pay to make the library a bit more widely usable, even if it will only affect a handful of users. For example ghc-mod depends on ghc and supports 7.10.3, so ghc-mod or anything that depends on it cannot be tested with hedgehog without this PR.

gwils:
I've just learnt that doctest also depends on the ghc library.
Hence, anyone who wants to support GHC 7.10.3 cannot use both doctest and hedgehog.
This is because doctest would force the boot library versions for the present GHC, which in the case of 7.10.3, would violate hedgehog's lower bound `transformers >= 0.5`

I would expect hedgehog wishes to cooperate well with other testing libraries, so I think this is now an easier choice.
I'm happy to maintain the CPP by the way, if you're concerned about maintenance going forward.

gwils:
[One of the jobs](https://travis-ci.org/hedgehogqa/haskell-hedgehog/jobs/336567011) failed because it went for 10 minutes without output while running the tests. This seems to happen occasionally with both the mafia-based travis setup and the new-build based one.

Over in #149 I've fixed it by inserting a call to `travis_wait`. For now, if you restart that specific job, it should hpoefully go green.

thumphries:
Have rebased this and will merge on green

gwils:
I've added a commit to set INSTALLED back to true in the travis config, so travis should now verify that we work with the ghc boot library versions of dependencies.

PR #151, opened 2018-02-04
Bump upper bounds for template-haskell and resourcet

thumphries:
- `template-haskell` is at `2.13` in GHC 8.4
- `resourcet` `1.2.0` was released the other day

We should be good for 8.4 after this, will aim for a point release

@nhibberd @gwils


gwils:
This looks good to me. I'd love if we could get #150 in before the point release too.

nhibberd:
:+1:

Issue #153, opened 2018-02-06
Unicode generator can generate invalid characters '\65534' and '\65535'


johnchandlerburnham:
'\65534' (U+FFFE) and '\65535' (U+FFFF) are listed as ["invalid characters"](http://www.charbase.com/fffe-unicode-invalid-character). Generating these with [Test.Hedgehog.Gen.unicode]( http://hackage.haskell.org/package/hedgehog-0.5.2/docs/src/Hedgehog-Internal-Gen.html#unicode) sets up a mismatch with other UTF8 libraries like Data.ByteString.UTF8.

```
> import qualified Data.ByteString.UTF8 as BU                  
BU> BU.toString (BU.fromString "\65534") == "\65533"          
True
BU> BU.toString (BU.fromString "\65535") == "\65533"          
True
```
 Arguably, [Data.ByteString.UTF8.fromString](https://hackage.haskell.org/package/utf8-string-1.0.1.1/docs/src/Data-ByteString-UTF8.html#toString) should do something different here. It might also make sense to have these characters only be generated by Test.Hedgehog.Gen.unicodeAll.

thumphries:
Thanks for the report!

This looks like a bug to me. We already filter some invalid unicode characters, and should probably be filtering these too

thumphries:
https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Internal/Gen.hs#L1042

thumphries:
Actually, given these are guaranteed not to be a valid unicode character, I wonder if they should be removed from `unicodeAll` too

thumphries:
Spoke to @jystic about this yesterday.

- `unicode` should not generate these characters
- `unicodeAll` should not be changed. Despite the name it is supposed to represent anything that can possibly appear in `Char` or `Text`. These values are representable so the most general generator should continue produce them.

erikd:
Since PR #154 is merged this can be closed right?

PR #154, opened 2018-02-07
unicode no longer generates noncharacters

johnchandlerburnham:
I haven't fully tested this, but I think this should work, right?

https://github.com/hedgehogqa/haskell-hedgehog/issues/153


thumphries:
Thanks for the PR!

If you could move the filter to `unicode` instead of `unicodeAll`, we can merge this.

See this comment - https://github.com/hedgehogqa/haskell-hedgehog/issues/153#issuecomment-363941585

thumphries:
Thanks!

@nhibberd eyes on this if you have time

moodmosaic:
:+1:

PR #158, opened 2018-02-14
Copy and paste LICENSE rather than using symlinks

gwils:
I've also updated the copyright years.


thumphries:
Why remove the symlinks? They're only there so that Hackage releases get the appropriate license, and `cabal sdist` follows the link just fine

gwils:
It's breaking my nix tooling when I `fetchFromGithub` and then try to `callCabal2nix`. Hedgehog doesn't officially support nix so this is more of a favour than anything.
I've talked to Jacob about this and he is cool with it.

thumphries:
Cool, sorry for the pushback, was missing context

gwils:
Thanks!

Issue #41, opened 2017-04-11
Documentation on Range.linearFrom parameters


chris-martin:
> `linearFrom :: Integral a => a -> a -> a -> Range a`

It isn't entirely clear to me what the first parameter does. Is it something like this?

```haskell
linearFrom :: Integral a 
  => a -- ^ Value to shrink toward (the value produced when the size parameter is zero)
  -> a -- ^ Lower bound
  -> a -- ^ Upper bound
  -> Range a
```

```haskell
data Range a =
  Range !a -- ^ Value to shrink toward (the value produced when the size parameter is zero)
        (Size -> (a, a))
```

jystic:
Yeah that's right, it's the value to shrink towards

PR #162, opened 2018-02-22
Add some documentation bits.

fisx:
I am open to suggestions re. style.  I just hacked in what would have helped me.  (-:

(That's one commit; the other is just cut & paste from #41.)

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/HTraversable.hs
+++ b/hedgehog/src/Hedgehog/Internal/HTraversable.hs
@@ -7,5 +7,26 @@ module Hedgehog.Internal.HTraversable (
 
 -- | Higher-order traversable functors.
 --
+-- This is used internally to make symbolic variables concrete given an 'Environment'.
+--
+-- The type checker will ask you to write 'HTraversable' instances for your action types.  If you do
+-- the trivial thing:
+--
+-- @
+-- data YourAction (v :: * -> *) = YourAction (Var YourType v)
+--   deriving (Eq, Show)
+--
+-- instance HTraversable YourAction where
+--   htraverse _ (YourAction var) = pure (YourAction var)  -- wrong!
+-- @
+--
+-- The type checker will complain that it @couldn't match type ‘g’ with ‘h’@.  To fix this, you need
+-- to 'htraverse' into your structure like so:
+--
+-- @
+-- instance HTraversable YourAction where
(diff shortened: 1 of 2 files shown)


hedgehog/src/Hedgehog/Internal/Range.hs
```diff
@@ -210,7 +211,11 @@ linear x y =
 --   >>> bounds 99 $ linearFrom 0 (-10) 20
 --   (-10,20)
 --
-linearFrom :: Integral a => a -> a -> a -> Range a
+linearFrom :: Integral a
+  => a -- ^ Value to shrink toward (the value produced when the size parameter is 0)
+  -> a -- ^ Lower bound (the bottom of the range when the size parameter is 99)
+  -> a -- ^ Upper bound (the top of the range when the size parameter is 99)
+  -> Range a
```
moodmosaic: 😅 Nitpick: Most (if not all?) comments in this codebase end with a `.`— It'd be nice if we do the same also here?

fisx: sure, i can fix all that (give me a few days).  anything else?

fisx: ah, travis.  i'll look into that, too...

fisx:
Ok, I hope this is a bit better.

I am aware that my style is odd and different from the rest of the documentation.  I'm happy to try harder, please let me know if i should!  (-:

fisx:
travis-timeout.  i'll push an empty commit to try again...

(please let me know if you would like me to squash anything / everything.)

hedgehog/src/Hedgehog/Internal/HTraversable.hs
```diff
@@ -7,5 +7,26 @@ module Hedgehog.Internal.HTraversable (
 
 -- | Higher-order traversable functors.
 --
+-- This is used internally to make symbolic variables concrete given an 'Environment'.
+--
+-- The type checker will ask you to write 'HTraversable' instances for your action types.  If you do
```
thumphries: I think from this line down might be better suited to a tutorial than the documentation.

If you could peel it off for now I would be happy to merge

thumphries:
Thanks!

fisx:
thank you!  (-:

PR #163, opened 2018-02-27
Generators for UTCTime and ZonedTime

amrhassan:
Hello,

I've found those generators useful to me and thought I would share them. I had to add dependencies on the `time` and `tz` packages. The former is where the data structures of `UTCTime` and `ZonedTime` come from and the latter enumerates all the time zones.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -72,6 +72,8 @@ library
     , transformers                    >= 0.4        && < 0.6
     , transformers-base               >= 0.4        && < 0.5
     , wl-pprint-annotated             >= 0.0        && < 0.2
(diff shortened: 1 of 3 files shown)


thumphries:
Hey, thanks for the PR!

These are a little bit out of Hedgehog's scope for dependency reasons. They belong in a separate package. Feel free to make one if I don't get there first.

Hope you're getting great value out of Hedgehog.

frasertweedale:
Hello!  I see currently release of *hedgehog* depends on *time*.  Maybe that happened more recently than this PR.  In any case, there would be no new dependencies to include generators for *time* types.  Would you be open to it now?

jappeace:
that's an older discussion, and yes please add these :sweat_smile:

PR #167, opened 2018-03-08
increase upper version bound of async and lifted async

taojang:
At work we are using this [library](https://github.com/mtesseract/nakadi-client/) which has following dependencies:
```
  - async-2.2.1
  - lifted-async-0.10.0
```

We have increased the upper version bounds of those two and run the test locally and they still pass.

Thank you very much in advance!


taojang:
sorry, it's an duplicate of https://github.com/hedgehogqa/haskell-hedgehog/pull/165

Issue #97, opened 2017-07-11
Top-level test runner


erikd:
Would be nice to have something like [disorderMain](https://github.com/ambiata/disorder.hs/blob/72309ceaec55a47d0d2577fa3ab12f087e377ddd/disorder-core/src/Disorder/Core/Main.hs#L16) in Hedgehog. 

Basically just something like:
```
runTests :: [IO Bool] -> IO ()
runTests tests = do
  hSetBuffering stdout LineBuffering
  hSetBuffering stderr LineBuffering
  results <- sequence tests
  unless (and results)
    exitFailure
```

thumphries:
> Does hedgehog have something like disorderMain?

Nope. Maybe it should? I've written the function above a dozen times now.

Alternative is to emphasise use through Tasty or HSpec, Hedgehog's test runner is arguably a bit better and I'd rather use it directly.

chris-martin:
As somebody involved in writing a lot of beginner Haskell tutorials, I'd really like to see this happen. We'd like to use Hedgehog as our standard testing recommendation for beginners, but `sequence`, `exitFailure`, and line buffering are a lot to have to explain upfront before you can even get started. We can't ask people to write a test runner that uses `sequence` while we're still teaching about `fmap`. Having to recommending a second test package to get the runner also seems overly cumbersome; do you not desire Hedgehog to be quickly usable "out of the box," so to speak?

thumphries:
> Perhaps support for Hedgehog with the 'detailed' Cabal testing interface is what we actually need.

`detailed` is not widely used, the community mostly just continued using `hspec` / `tasty`. So I wouldn't put too much weight on that, though supporting it might encourage use. Last I looked, `cabal new-test` didn't support it yet! It might now, not sure.

It's not that we want to screw over `teh beginners`; we spent quite a while trying to settle on the right code to include, but every choice we made seemed adhoc and surprising, so we slept on it.

In general we've pushed adhoc experiments out of `hedgehog` proper, like `hedgehog-fn`. Packages that prove truly indispensable can be brought back into the main library if necessary, though we prefer to distribute the maintenance burden. A second library that we link to in documentation is probably the only resolution for this one.

> do you not desire Hedgehog to be quickly usable "out of the box," so to speak?

> We can't ask people to write a test runner that uses `sequence` while we're still teaching about `fmap`

[...]

PR #168, opened 2018-03-11
Add a runTests function

erikd:
And actually use it in the Hedhehog tests.

Closes: https://github.com/hedgehogqa/haskell-hedgehog/issues/97

@thumphries @jystic @nhibberd

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -17,6 +17,9 @@ module Hedgehog.Internal.Runner (
   , checkSequential
   , checkGroup
 
+  -- * Top level testsuite runner
+  , runTests
+
   -- * Internal
   , checkReport
   , checkRegion
@@ -25,6 +28,7 @@ module Hedgehog.Internal.Runner (
 
 import           Control.Concurrent.STM (TVar, atomically)
 import qualified Control.Concurrent.STM.TVar as TVar
+import           Control.Monad (unless)
 import           Control.Monad.Catch (MonadCatch(..), catchAll)
 import           Control.Monad.IO.Class (MonadIO(..))
 
@@ -46,8 +50,11 @@ import           Hedgehog.Range (Size)
 
 import           Language.Haskell.TH.Lift (deriveLift)
 
+import           System.Exit (exitFailure)
 #if mingw32_HOST_OS
-import           System.IO (hSetEncoding, stdout, stderr, utf8)
+import           System.IO (BufferMode (LineBuffering), hSetBuffering, hSetEncoding, stdout, stderr, utf8)
+#else
+import           System.IO (BufferMode (LineBuffering), hSetBuffering, stderr, stdout)
 #endif
 
 -- | Configuration for a property test run.
@@ -401,6 +408,16 @@ checkParallel =
       , runnerVerbosity =
           Nothing
       }
+
+-- | Like `runTests` but exit with `exitFailure` if one or more of the tests
+-- fail.
+runTests :: [IO Bool] -> IO ()
+runTests tests = do
+  hSetBuffering stdout LineBuffering
+  hSetBuffering stderr LineBuffering
+  result <- and <$> sequence tests
+  unless result
+    exitFailure
 
 ------------------------------------------------------------------------
 -- FIXME Replace with DeriveLift when we drop 7.10 support.
--- a/hedgehog/test/test.hs
+++ b/hedgehog/test/test.hs
@@ -1,20 +1,12 @@
-import           Control.Monad (unless)
-import           System.IO (BufferMode(..), hSetBuffering, stdout, stderr)
-import           System.Exit (exitFailure)
+import           Hedgehog (runTests)
 
 import qualified Test.Hedgehog.Seed
 import qualified Test.Hedgehog.Text
 
 
 main :: IO ()
-main = do
-  hSetBuffering stdout LineBuffering
-  hSetBuffering stderr LineBuffering
-
-  results <- sequence [
+main =
+  runTests [
       Test.Hedgehog.Text.tests
     , Test.Hedgehog.Seed.tests
     ]
-
-  unless (and results) $
-    exitFailure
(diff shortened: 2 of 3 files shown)


hedgehog/test/test.hs
```diff
@@ -1,18 +1,10 @@
...
 main :: IO ()
-main = do
-  hSetBuffering stdout LineBuffering
-  hSetBuffering stderr LineBuffering
-
-  results <- sequence [
-      Test.Hedgehog.Text.tests
+main =
+  runTests
+    [ Test.Hedgehog.Text.tests
```
nhibberd: Probably good to keep the same syntax style here, start of list on previous line that is

erikd: :+1: Fixed the formatting in the comment for `runTests` as well.

thumphries:
I feel a bit weird exporting a function that calls `exitWith` from the top level, especially with an innocuous name like `runTests`. Wondering if it would make more sense from `Hedgehog.Main` or something like that? Or a name with `suite` or `main` in it?

(not requesting changes per se, just wondering how you two feel about it)

erikd:
I'm not married to the name, I just don't want to have to redefine this function in every test suite I write. I also think what you said makes a lot of sense. So how about:

* We rename it `runTestsuie`.
* Move it to `Hedgehog.Main` and not export if from `Hegdehog` requiring people to explicitly import `Hedgehog.Main`.
* Since if calls `exitFailure` on any failure, maybe it should also call `exitSuccess` on success. Ie this function *always* terminates.

erikd:
Hmm, no that I think about it, I'm mot so sure about the last part. I could imagine wanting to do:
```
   runTests
        [ Test.Wibble.One
        ]
  -- Only run the following tests if the previous one passed.
   runTests
        [ Test.Wibble.Two
        , Test.Wibble.Three
        ]
```

moodmosaic:
>Move it to Hedgehog.Main and not export if from Hegdehog requiring people to explicitly import Hedgehog.Main.

Hmm... Isn't this a little bit weird? Perhaps it could be exposed as 'internal' from Hedgehog itself so we keep importing only the essential stuff:

```haskell
import           Hedgehog
```

erikd:
The idea of putting this in a separate `Hedgehog.Main` is growing on me. Maybe we should also provide more than one runner. `Hedgehog.Main`might provide fhe functions:

```
setLineBufferring :: IO ()

-- Run the tests and return `True` if they all passed, `False` otherwise.
runTestsuite :: [IO Bool] -> IO Bool

-- If any test fails, exit with `ExitFailure`.
runTestsuiteExitOnFail :: [IO Bool] -> IO ()

-- This _always_ exits, `ExitFailure` if any test fails, `ExitSuccess` otherwise.
runTestsuiteExit :: [IO Bool] -> IO a

```

moodmosaic:
Are we 100% sure about the `Main` part? Not that I can currently come up with an alternative...

erikd:
@moodmosaic The import you have in your example are the ones you would put in an actual test suite.. I usually keep actuall tests out of the top level test runner. With this new functionality, my top level test runner would look like:

```
import qualified Hedgehog.Main as HMain

import qualified Test.Wibble.One
import qualified Test.Wibble.Two
import qualified Test.Wibble.Three

main :: IO ()
main = do
  HMain.setLineBuffering
  HMain.runTestsuiteExit
    [ Test.Wibble.One
    , Test.Wibble.Two
    , Test.Wibble.Three
    ]
   
I'm not married to the name `Main`. Might be `Run` or `Runner` instead. Or maybe with these new names and type signatures we can just export them from `Hedgehog`.

**Comment by gwils** [comment]
I don't like the name Hedgehog.Main, but Run or Runner I could get behind. It just feels a bit wrong to me for a library (not an executable) to export a module called Main. I would prefer that this be exported from Hedgehog rather than a separate module regardless.

My preference is the version that calls `exitFailure` on failure but _does not_ call `exitSuccess` on success. As Erik has pointed out, you may wish to call many test runs sequentially.

**Comment by erikd** [comment]
> My preference is the version that calls exitFailure on failure but does not call exitSuccess on success.

This is why my (strawman) proposal now has three versions of the runner function :).

File: hedgehog/src/Hedgehog/Internal/Runner.hs [hunk 2]
```diff
@@ -394,6 +401,35 @@ checkParallel =
           Nothing
       }
 
+-- | Run list of tests.
+--
+-- Given a number of separate modules all defining test funtion named `tests`,
+-- the individual module level tests can be collected together and run in
+--  sequence (one module of tests at a time) and the funcion will will exit
```
moodmosaic: Nitpick: there seems to be an extra space between `sequence`.
Nitpick: `will` occurs two times in a row on that sentence.

erikd: Yeah, this PR will not go ahead as it is anyway.

thumphries:
Closing this based on Slack discussions we had back in March.

The decision was to not rush into exporting a top-level runner, since public exports are forever and we all had some reservations about the current design. There is also the question of terminal encoding, Windows, etc to consider.

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -394,6 +401,22 @@ checkParallel =
...
+  fmap and . sequence
+
+-- | Like `runTestsuite` but exit with `exitFailure` if one or more of the
+-- tests fail.
+
+runTestsExitFail :: [IO Bool] -> IO ()
+runTestsExitFail tests = do
+  result <- runTests tests
+  unless result
+    exitFailure
```
jystic: If you rename this to `runTests` and delete the other function I'll be happy

hedgehog/test/test.hs
```diff
@@ -1,18 +1,10 @@
...
-import           System.Exit (exitFailure)
+import           Hedgehog (runTestsExitFail)
 
 import qualified Test.Hedgehog.Text
 
 
 main :: IO ()
-main = do
-  hSetBuffering stdout LineBuffering
-  hSetBuffering stderr LineBuffering
```
jystic: Don't we need this? Should probably be in `runTests`

https://github.com/ambiata/disorder.hs/blob/master/disorder-core/src/Disorder/Core/Main.hs#L26-L28

jystic:
💯

PR #169, opened 2018-03-12
.travis.yml: DIsallow failures with ghc 8.4

erikd:
Ghc 8.4 has now had an official release so the build should succeed.


erikd:
Looks like hashtables doesn't build with 8.4 yet.

gwils:
If you run this again, it should work now.
See https://github.com/gregorycollins/hashtables/pull/47

thumphries:
Thanks Erik!

PR #179, opened 2018-04-15
Stack.yaml and Makefiles

bitemyapp:
Was getting ready to do some work on the library, this'll spare anybody else needing to do the same prep work.


thumphries:
Sorry, not going to merge this. Hedgehog should always build out of the box with `stack init --solver`. Beyond that we don't want any stack-related PR noise.

PR #196, opened 2018-05-15
add relation stuff ala === and /==

chessai:
This PR adds 4 functions, that have similar behaviour to === and /==, except they correspond to different relations.

(<==) ~> (<=)
(>==) ~> (>=)

(>>>) ~> (>)
(<<<) ~> (<)

As for naming, I tried to keep them to three characters while making sure their names didn't conflict with anything popular.

The reason I wanted this is that I was doing some numerical computations and wanted to test that the result of a computation was less than or equal to a particular value.

### DIFF
--- a/hedgehog-example/src/Test/Example/Basic.hs
+++ b/hedgehog-example/src/Test/Example/Basic.hs
@@ -70,8 +70,8 @@ prop_foo =
 
     guard (y `mod` 2 == (1 :: Int))
 
-    assert $
-      y < 87 && x <= 'r'
+    diff y (<) 87
+    diff x (<=) 'r'
 
 ------------------------------------------------------------------------
 -- Example 3
@@ -251,7 +251,7 @@ prop_record =
   property $ do
     x <- forAll genRecord
     y <- forAll genRecord
-    x === y
+    diff x (==) y
 
 prop_different_record :: Property
 prop_different_record =
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -56,6 +56,10 @@ module Hedgehog.Internal.Property (
   , assert
   , (===)
   , (/==)
+  , (<==)
+  , (>==)
+  , (<<<)
+  , (>>>)
 
   , eval
   , evalM
@@ -577,6 +581,50 @@ infix 4 /==
           "━━━ Both equal to ━━━"
         , showPretty x
         ]
+
+infix 4 <==
+
+-- | Fails the test if the right argument is less than the left.
+--
+(<==) :: (MonadTest m, Ord a, Show a, HasCallStack) => a -> a -> m ()
+(<==) x y = do
+  ok <- withFrozenCallStack $ eval (x <= y)
+  if ok then
+    success
+  else
+    withFrozenCallStack $ failDiff x y
+
+infix 4 >==
+
+-- | Fails the test if the right argument is greater than the left.
+(>==) :: (MonadTest m, Ord a, Show a, HasCallStack) => a -> a -> m ()
+(>==) x y = do
+  ok <- withFrozenCallStack $ eval (x >= y)
+  if ok then
+    success
+  else
+    withFrozenCallStack $ failDiff x y
+
+infix 4 <<<
+
+-- | Fails the test if the right argument is less than or equal to the left.
+(<<<) :: (MonadTest m, Ord a, Show a, HasCallStack) => a -> a -> m ()
+(<<<) x y = do
+  ok <- withFrozenCallStack $ eval (x < y)
+  if ok then
+    success
+  else
+    withFrozenCallStack $ failDiff x y
+
+infix 4 >>>
+
+(>>>) :: (MonadTest m, Ord a, Show a, HasCallStack) => a -> a -> m ()
+(>>>) x y = do
+  ok <- withFrozenCallStack $ eval (x > y)
+  if ok then
+    success
+  else
+    withFrozenCallStack $ failDiff x y
 
 -- | Fails the test if the value throws an exception when evaluated to weak
 --   head normal form (WHNF).
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -54,12 +54,9 @@ module Hedgehog.Internal.Property (
   , failure
   , success
   , assert
+  , diff
   , (===)
   , (/==)
-  , (<==)
-  , (>==)
-  , (<<<)
-  , (>>>)
 
   , eval
   , evalM
@@ -157,7 +154,6 @@ newtype PropertyT m a =
     , MonadState s
     , MonadError e
     )
-
 -- NOTE: Move this to the deriving list above when we drop 7.10
 deriving instance MonadResource m => MonadResource (PropertyT m)
 
@@ -326,9 +322,6 @@ instance Monad m => Monad (TestT m) where
       unTest m >>=
       unTest . k
 
-  fail =
-    Fail.fail
-
 instance Monad m => MonadFail (TestT m) where
   fail err =
     TestT . ExceptT . pure . Left $ Failure Nothing err Nothing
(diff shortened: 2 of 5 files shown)


andrewthad:
The names `>>>` and `<<<` conflict with some stuff from `Control.Arrow`. I find those names confusing, but I like the idea here. Maybe non-infix names would be better here, but that's just my two cents as a very casual user of `hedgehog`. The project maintainers probably have better vision for this kind of thing.

chessai:
I considered Control.Arrow; but I weigh that pretty low, since it isn't used much. Other people might weigh that higher.

As far as the names being confusing, I agree that they definitely may be. I considered (<<=)<sub>hedgehog_pr</sub>, but that looks too close to (=<<)<sub>base</sub>, and its counterpart (>>=)<sub>hedgehog_pr</sub> is nominally (in the linguistic sense) equivalent to (>>=)<sub>base</sub>. If anyone has any idea for a better name for those, I certainly wouldn't object. I don't know if it's necessary to continue the trend of three characters - it's possible that they could be english character names.

thumphries:
They’re also used in Control.Category, so I will also suggest we find different names or operators.

We have not really thought through how many base relations to include as special Hedgehog functions or operators. Now’s probably a good time to think about it.

My gut feeling is that it’s worthwhile wherever we can provide a superior counterexample report by doing so.

chessai:
@thumphries Would you be in favour of a non-infix operator name?

chessai:
ping

jystic:
Not a fan of this I'm sorry, I'd like to keep operators to a minimum. I think this is prettier anyway:
```
assert $ x > y
```
Of course it doesn't do the diff, but that's mainly useful for large data structures and I wouldn't have though comparing those for ordering is common enough to include dedicated functions. Maybe you have an example?

Also there's no way I'd merge operators which clash with both `Control.Category` and `Control.Arrow` (i.e. `>>>` and `<<<`)

jystic:
I apologize for the tone of the last message I think I was in a bit of a bad mood when I wrote it.

I've been considering this PR since I last replied, and I think I found something I'm happy with. I pushed an update to to your branch which renames the operators to use an `@` prefix which I pronounce "assert" and then the standard haskell name for the operator. I kind of like this even though it's a bit ugly because once you learn the rule (prefix with `@`) then you know how to get all the rest.

jystic:
As discussed with @amosr @HuwCampbell, I've added function to assert relations directly.

Arguably the operators are no longer required, but they do have the advantage of perhaps slightly nicer failure output. Any opinions either way on whether we should keep the operators given the new `relation` function is appreciated.

### Failure using operators
Screenshot 2019-04-17 at 6 52 22 PM

### Failure using `relation`
Screenshot 2019-04-17 at 6 53 29 PM

moodmosaic:
Keeping the function that @jystic added should be good enough, perhaps I'd consider giving it a different name (I think I've actually never seen `relation` as a function name in any testing tool out there).

Basically we have this already which is good enough

```
assert $ x > y
```

perhaps the proposed `relation` one could be renamed, e.g. → `relate`, but really not sure about adding more operators :confused: 

---

The issue I see with adding more operators is that it makes it hard(er) to draw a line; where do we stop adding more *assertion*-specific stuff?

I often see assertions packed on separate libraries (see [Groovy Power Assert](https://dontmindthelanguage.wordpress.com/2009/12/11/groovy-1-7-power-assert/), [Unquote](https://github.com/SwensenSoftware/unquote), etc). It's just the Unix philosophy at the end; do *one* thing and do it well...

jystic:
>  think I've actually never seen `relation` as a function name in any testing tool out there

That is what it's called though https://en.wikipedia.org/wiki/Relational_operator

moodmosaic:
>That is what it's called though https://en.wikipedia.org/wiki/Relational_operator

That's great then :+1:

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -550,24 +574,85 @@ assert b = do
...
+--   /Synonym for '@/='./
+--
 (/==) :: (MonadTest m, Eq a, Show a, HasCallStack) => a -> a -> m ()
-(/==) x y = do
+(/==) x y =
+  withFrozenCallStack $ x @== y
+
+infix 4 @==
+
+-- | Fails the test if the two arguments provided are equal.
```
charleso: "not equal"? (And I think the `@/=` comment is the wrong way around too)

jystic: yeah, you're right oops

i think i might remove the operators anyway after chatting to a few people about this PR

hedgehog-example/src/Test/Example/Basic.hs
```diff
@@ -70,8 +70,17 @@ prop_foo =
...
+    diff x (<=) 'r'
+
+    comparison y (<) 87
+    comparison x (<=) 'r'
+
+    relation y (<) 87
+    relation x (<=) 'r'
+
+    relate y (<) 87
+    relate x (<=) 'r'
```
jystic: Would appreciate some comments on which name people like best or some other one? The main point of this is to provide a diff, otherwise you'd just use assert and annotate.

Also is the infix style crazy? it's kind of non-haskellish but it does read pretty nice.

chessai:
> I apologize for the tone of the last message I think I was in a bit of a bad mood when I wrote it.

I didn't pick up on any rudeness, so it's ok.

Thanks for picking up some work on the branch

hedgehog-example/src/Test/Example/Basic.hs
amosr: no strong opinions, but `comparison` is my favourite. I like the infix style, I think it makes the meaning pretty obvious.

moodmosaic: I like `diff` and `relate`, I like how they sound, *diff this with that*, *relate this with that*.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -470,8 +471,8 @@ writeLog x =
 --   combinators.
 --
 failWith :: (MonadTest m, HasCallStack) => Maybe Diff -> String -> m a
-failWith diff msg =
-  liftTest $ mkTest (Left $ Failure (getCaller callStack) msg diff, [])
+failWith mdiff msg =
```
moodmosaic: Do we care to pick up a better name instead of `mdiff`? Although I think we do similar in other places as well so probably not a huge deal.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -550,33 +561,52 @@ assert b = do
...
+
+comparison :: (MonadTest m, Show a, Show b, HasCallStack) => a -> (a -> b -> Bool) -> b -> m ()
+comparison = diff
+
+relation :: (MonadTest m, Show a, Show b, HasCallStack) => a -> (a -> b -> Bool) -> b -> m ()
+relation = diff
+
+relate :: (MonadTest m, Show a, Show b, HasCallStack) => a -> (a -> b -> Bool) -> b -> m ()
+relate = diff
+
```
moodmosaic: See comment at https://github.com/hedgehogqa/haskell-hedgehog/pull/196/files#r276932847.

moodmosaic: I think `diff` sounds more clear than all the others, short and sweet, *diff this with that* :rocket: :rocket: :rocket:

hedgehog/src/Hedgehog/Internal/Property.hs
jystic: I tent to use an `m` prefix for maybe things is all. So when you peel it off it looks like this:
```haskell
foo mx =
  case mx of
    Nothing ->
      ..
    Just x ->
      ..
```

hedgehog-example/src/Test/Example/Basic.hs
jystic: One thing @moodmosaic and I discussed with `comparison` is that it's not a verb, like if we could use `compare` that would be awesome but it's already taken by `Ord`

jystic:
Looks like `diff` is the winner 📜

I am still willing to change it if someone has better ideas / good arguments as long as it's prior to the next release.

moodmosaic:
:+1:

hedgehog/src/Hedgehog/Internal/Tripping.hs
```diff
@@ -3,9 +3,9 @@ module Hedgehog.Internal.Tripping (
     tripping
   ) where
 
-import           Hedgehog.Internal.Property
-import           Hedgehog.Internal.Show
-import           Hedgehog.Internal.Source
+import           Hedgehog.Internal.Property (MonadTest, Diff(..), success, failWith)
+import           Hedgehog.Internal.Show (valueDiff, mkValue, showPretty)
+import           Hedgehog.Internal.Source (HasCallStack, withFrozenCallStack)
```
moodmosaic: Good catch :+1:

PR #197, opened 2018-05-20
Seed: Fix compiler warning

erikd:
`Int32` is only used on 32 bit platforms and `Int64` is only used on 64 bit platforms so tweak the `CPP` stuff so that there are no warnings (mafia made me do this).

Built this in 32 bit Debian as a test. Zero warnings. Tests passed.


thumphries:
:+1:

PR #200, opened 2018-05-26
Property: Add evalLeft and isLeft

erikd:
I am using `isLeft` now for the use case mentioned in the haddock text.

`evalLeft` I am less sure of. I can see it being useful for testing that correct errors are being generated. An example would be a generator that generates malformed binary serialisations and the expected error message.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -61,7 +61,9 @@ module Hedgehog.Internal.Property (
   , evalM
   , evalIO
   , evalEither
+  , evalLeft
   , evalExceptT
+  , isLeft
 
   -- * Internal
   -- $internal
@@ -79,7 +81,7 @@ module Hedgehog.Internal.Property (
   ) where
 
 import           Control.Applicative (Alternative(..))
-import           Control.Monad (MonadPlus(..))
+import           Control.Monad (MonadPlus(..), void)
 import           Control.Monad.Base (MonadBase(..))
 import           Control.Monad.Catch (MonadThrow(..), MonadCatch(..))
 import           Control.Monad.Catch (SomeException(..), displayException)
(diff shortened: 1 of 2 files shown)


hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -604,6 +606,27 @@ evalEither = \case
...
+-- | Fails the test if the 'Either' is 'Right'.
+-- An example use case would be testing that a parser the returns an 'Either'
+-- fails on all invalid inputs.
+--
+isLeft :: (MonadTest m, Show a, HasCallStack) => Either x a -> m ()
+isLeft = \case
+  Left _ ->
+    pure ()
+  Right x ->
+    withFrozenCallStack $ failWith Nothing $ showPretty x
```
moodmosaic: Wondering if this duplication tries to tell us something :confused: 

image

erikd: Updated the PR with more succinct versions.

moodmosaic:
>Updated the PR with more succinct versions.

Technically, it looks good :+1:, but I think the duplication in the [original form](https://github.com/hedgehogqa/haskell-hedgehog/pull/200#discussion_r191042462) was clearly showing that these props, although probably useful, are fairly trivial.

By fairly trivial I mean that it's probably a whole lot easier to just define them on an ad-hoc basis, because once they're part of the public API they stay there forever (kind of).

Once we start taking those and similar ones, others might start adding more and more, and [we've been in a similar situation with the F# version](https://github.com/hedgehogqa/fsharp-hedgehog/issues/139#issuecomment-338966392) where we [ended up putting them in a contrib](https://github.com/cmeeren/fsharp-hedgehog-experimental/issues/30) project outside of org...

That being said, and after some careful consideration, I think we should close this. /cc @jystic @thumphries

thumphries:
Agree with Nikos here - these are a bit too trivial to justify breaking with the existing naming conventions / library aesthetic. We need to take care not to let the core library become a bucket of utility code.

There's probably room for a `hedgehog-either` package, since there are many potentially useful combinators for Either and ExceptT.

PR #202, opened 2018-06-12
Add note about enumBounded

thumphries:
'enumBounded' is partial in some cases. Usually you actually want something like `Gen.linear minBound maxBound` or similar. The error message is pretty inscrutable when you mess it up (no source locations), so I think a cautionary message is worthwhile.

Bikeshed on wording welcome

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -957,6 +957,8 @@ enum lo hi =
 -- enumBounded :: 'Gen' 'Bool'
 -- @
 --
+--   /This is implemented in terms of the 'Enum' class, and thus may be
+--    partial for integral types larger than 'Int', e.g. 'Word64'./
 enumBounded :: (MonadGen m, Enum a, Bounded a) => m a
 enumBounded =
   enum minBound maxBound


jystic:
Maybe we can do a check and give an `error` which is more sensible?

moodmosaic:
I think we need both―a note in the comments, so it shows up in the docs, and also a check with an error message (in the error message we can put that same stuff as in the notes).

moodmosaic:
Should be OK to merge this as-is and have the check on a different PR, if that's easier.

moodmosaic:
>Bikeshed on wording welcome

Hmm, perhaps we should mention what the docs say for [`fromEnum`](http://hackage.haskell.org/package/base-4.11.1.0/docs/Prelude.html#v:fromEnum) as well?

>Convert to an Int. It is implementation-dependent what fromEnum returns when applied to a value that is too large to fit in an Int.

jystic:
I'm not sure the `fromEnum` docs add that much? it kind of implies that it might actually work when it probably won't, it is "more correct" but I think just exploding is probably what is going to happen.

moodmosaic:
Because it's `fromEnum` that isn't total at the end:

```diff
diff --git a/hedgehog/test/Test/Hedgehog/Text.hs b/hedgehog/test/Test/Hedgehog/Text.hs
index 0a11e91..a320100 100644
--- a/hedgehog/test/Test/Hedgehog/Text.hs
+++ b/hedgehog/test/Test/Hedgehog/Text.hs
@@ -29,7 +29,7 @@ genOdd =

 genSeed :: Gen Seed
 genSeed =
-  Seed <$> Gen.word64 Range.constantBounded <*> fmap fromIntegral genOdd
+  Seed <$> Gen.enumBounded <*> fmap fromIntegral genOdd

 genPrecedence :: Gen Int
 genPrecedence =
```

```
[thread killed]
test.exe:
  Enum.fromEnum{Word64}:
    value (18446744073709551615) is outside of Int's bounds
      (-9223372036854775808,9223372036854775807)
```

But yeah you're probably right that we might not need to get in to such detail.

thumphries:
Will mess around and try to produce a better error.

Happy to kill the documentation if y'all prefer. I hadn't quite joined the dots that this had to be going via `fromEnum`, so I definitely overuse this function where I probably shouldn't.

jystic:
I think the error would be good, but no reason to hold up this comment.

Issue #191, opened 2018-05-13
mixGamma produces bad gammas


Lysxia:
The `mixGamma` function should "fix" its result if the popcount is too low, but in its current version it is doing the opposite.

https://github.com/hedgehogqa/haskell-hedgehog/blob/d97e846e7503b27a1ea68cef0a676fb13cd2ff97/hedgehog/src/Hedgehog/Internal/Seed.hs#L193-L198

This is a bug inherited from a typo in the code of the original SplitMix paper. The text gives the correct interpretation:

> Therefore we require that the number of such pairs, as computed by `Long.bitCount(z ^ (z >>> 1))`, exceed `24`; if it does not, then the candidate `z` is replaced by the `XOR` of `z` and `0xaaaaaaaaaaaaaaaaL`

I found that in this blogpost: http://www.pcg-random.org/posts/bugs-in-splitmix.html

This also affects `fsharp-hedgehog`.

moodmosaic:
>I found that in this blogpost: http://www.pcg-random.org/posts/bugs-in-splitmix.html

Is the author of that article (Melissa O’Neill) biased against SplitMix?

- [On Melissa O’Neill’s PCG random number generator](https://lemire.me/blog/2017/08/15/on-melissa-oneills-pcg-random-number-generator)
- [Hacker News](https://news.ycombinator.com/item?id=15019071)

Perhaps it's `n >= 24` according to the paper (#124), perhaps it's `n < 24` according to Oracle's JDK8, either way all statistical tests for measuring the quality of that random number generator pass.

```bash
# n >= 24 (commit e187c131f76619e53d29c8aee9dddb9016bedb71)
$ dieharder-input | dieharder -a -g 200
```

<details>
  <summary>Test results (Click to expand)</summary>
<pre>
#=============================================================================#
#            dieharder version 3.31.1 Copyright 2003 Robert G. Brown          #
#=============================================================================#
   rng_name    |rands/second|   Seed   |
stdin_input_raw|  3.73e+06  |3112223130|
#=============================================================================#
        test_name   |ntup| tsamples |psamples|  p-value |Assessment
#=============================================================================#
   diehard_birthdays|   0|       100|     100|0.34563424|  PASSED
      diehard_operm5|   0|   1000000|     100|0.91055807|  PASSED
  diehard_rank_32x32|   0|     40000|     100|0.54739259|  PASSED
    diehard_rank_6x8|   0|    100000|     100|0.26505805|  PASSED
[...]

PR #207, opened 2018-07-02
Add tests to verify that our SplitMix/Seed avoids pathological γ-values

moodmosaic:
We were missing the test part in Haskell, which is pretty [similar to what we did in F#](https://github.com/hedgehogqa/fsharp-hedgehog/pull/161/commits/1746ddffd643a273662e277d50bec7f5bc024510). See Melissa E. O'Neill's post with title [Bugs in SplitMix(es)](http://www.pcg-random.org/posts/bugs-in-splitmix.html), #191, and 39b15b9b4d147f6001984c4b7edab00878269da7, for more context.

/cc @jystic @thumphries

### DIFF
--- a/hedgehog/test/Test/Hedgehog/Seed.hs
+++ b/hedgehog/test/Test/Hedgehog/Seed.hs
@@ -1,0 +1,84 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Test.Hedgehog.Seed (
+    tests
+  ) where
+
+import           Data.Foldable (for_)
+
+import           Hedgehog
+import qualified Hedgehog.Internal.Seed as Seed
+
+data Assert =
+  Assert {
+      expected :: !Seed
+    , actual   :: !Seed
+    } deriving (Show)
+
+-- | Verify that SplitMix avoids pathological γ-values, as discussed by
+--   Melissa E. O'Neill in the post with title Bugs in SplitMix(es), at
+--   http://www.pcg-random.org/posts/bugs-in-splitmix.html
+--
+--   See also:
+--   https://github.com/hedgehogqa/haskell-hedgehog/issues/191
+--
+prop_avoid_pathological_gamma_values :: Property
+prop_avoid_pathological_gamma_values =
+  withTests 1 . property $ do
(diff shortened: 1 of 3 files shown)


hedgehog/test/Test/Hedgehog/Seed.hs
```diff
@@ -0,0 +1,81 @@
...
+
+-- | Verify that SplitMix avoids pathological γ-values, as discussed by
+--   Melissa E. O'Neill in the post with title Bugs in SplitMix(es), at
+--   http://www.pcg-random.org/posts/bugs-in-splitmix.html
+--
+--   See also:
+--   https://github.com/hedgehogqa/haskell-hedgehog/issues/191
+--
+prop_avoid_pathological_gamma_values :: Property
+prop_avoid_pathological_gamma_values =
```
thumphries: This looks fully deterministic to me, is that the case?

If so, we don't need to run this 100 times.

Suggest something like 

```haskell
withTests 1 . property $ do
  for asserts $ \a ->
    expected a === actual a
```

moodmosaic: Indeed, those are deterministic, so just doing the comparison once is enough. Good catch 👍 Fixed and force-pushed.

moodmosaic:
Hmm, not sure why 1 of the 2 Travis CI checks fails. I think it's because I forced pushed the branch to the wrong origin, and then I forced pushed to the right one.

thumphries:
Thanks!

hedgehog/test/Test/Hedgehog/Seed.hs
```diff
@@ -0,0 +1,84 @@
...
+
+import           Data.Foldable (for_)
+
+import           Hedgehog
+import qualified Hedgehog.Internal.Seed as Seed
+
+data Assert =
+  Assert {
+      expected :: !Seed
+    , actual   :: !Seed
```
moodmosaic: Ugh! I forgot we're strict with no alignment anywhere (look at line 14). Happy to clean this up if you wish. /cc @jystic @thumphries

Issue #209, opened 2018-07-09
Get counterexample from Failure


japesinator:
I'm working on an application using hedgehog and attempting to do some further analysis on why a given test failed. To do this, I'd like to get a specific counterexample for a given property (not necessarily a `Property`, willing to manipulate the types if necessary).

I attempted to do this by looking into `Failure`s, but as far as I can tell the information on how something failed is represented as a `String`. 

Ideally, I'd like something `:: PropertyT m a -> m a` or similar. Is this possible? Happy to contribute code if necessary.

arianvp:
I had the exact same question today, and someone else seems to have done this with quickcheck using
type families:

https://github.com/nick8325/quickcheck-with-counterexamples

Should be possible to add it to Hedgehog too.

I have many datatypes for which I don't have Show instances (on purpose, for example `data Password`), which I still want to use in property-based testing, but currently I can't which is very frustrating.

thumphries:
I think this can be done at the `PropertyT` level, as you suggest.

If you add a type parameter to `Result`, you should be able to write a variant of `check` / `recheck` that produces an `a` on failure. We could probably do so without breaking any public APIs.

thumphries:
If you'd like to work on this, get in touch and we can add you to the Hedgehog Slack.

arianvp:
I have made some sketches for the API to see if this stuff is possible, and it seems it is!
Just needs some actualy coding!

We can keep all the existing machinery `GenT,  TestT

and then created an _indexed_ version of PropertyT

The nice thing is, that we can probably make a backwards compatible wrapper
around `IxPropertyT` that ignores all the extra state, and exposes the exact same old
interface.  Old users will not have breaking changes, and new users can actually get counter examples back!


```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE RebindableSyntax #-}
{-# LANGUAGE FlexibleInstances #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE StandaloneDeriving #-}
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE TypeOperators #-}

import qualified Prelude
import Prelude hiding ((>>=), return)
import Control.Monad
import Control.Applicative

-- | This is somewhere in Hedgehog already
checkIt :: TestT (GenT IO) () -> IO Bool
checkIt = undefined

class IxFunctor m where
[...]

PR #219, opened 2018-08-25
WIP: Allow for returning counterexamples

arianvp:
Fixes https://github.com/hedgehogqa/haskell-hedgehog/issues/209

This is still a WIP. but the first step is completed (adding a type parameter to Property)

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -127,10 +127,10 @@ import           Language.Haskell.TH.Lift (deriveLift)
 -- | A property test, along with some configurable limits like how many times
 --   to run the test.
 --
-data Property =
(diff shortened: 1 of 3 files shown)


arianvp:
Need some advice here.

Considering `TestT` is the actual thing that is returning a `Failure`, by means of an `ExceptT`  I actually think the type
of `PropertyT` needs to be more like:

```
newtype TestT e m a
newtype PropertyT e m a

data Property e = { PropertyT e m () }
```

reflecting that a `PropertyT` can return a counterexample `e`.

but that actually seems to be a bit of a bigger refactoring, as I'll need to refactor `MonadTest` to carry
around `e` too

but then functions like `eval`, `evalEither` and `evalM` become unimplementable...   so I'm kind of stuck and it becomes a bit of a larger refactoring than I would've hoped

PR #220, opened 2018-08-26
Update resourcet upper bound

osa1:
resourcet starts using `MonadUnliftIO` instead of `MonadBaseControl IO`
with 1.2 so hedgehog can only be used with `resourcet < 1.2`.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -64,7 +64,7 @@ library
     , pretty-show                     >= 1.6        && < 1.8
     , primitive                       >= 0.6        && < 0.7
     , random                          >= 1.1        && < 1.2
-    , resourcet                       >= 1.1        && < 1.3
+    , resourcet                       >= 1.1        && < 1.2
     , semigroups                      >= 0.16       && < 0.19
     , stm                             >= 2.4        && < 2.5
     , template-haskell                >= 2.10       && < 2.14


osa1:
Compare:

- http://hackage.haskell.org/package/resourcet-1.1.11/docs/Control-Monad-Trans-Resource.html#v:runResourceT
- http://hackage.haskell.org/package/resourcet-1.2.0/docs/Control-Monad-Trans-Resource.html#v:runResourceT

gwils:
`hedgehog` itself does not use `runResourceT`. It only depends on `resourcet` to provide `MonadResource` instances, which are compatible across `resourcet` 1.1 and 1.2 according to my testing.

`hedgehog-example` uses `runResourceT`, and it's upper bound is rightly `resourcet < 1.2`. So I don't think anything needs to be done here.

thumphries:
Feel free to reopen if George is wrong

osa1:
Hmm, so as long as I don't use `test . runResourceT` it's fine to use resourecet-1.3 and hedgehog in an application. I guess it's OK to leave the upper bound as-is. It's still confusing that the example in the documentation

```haskell
  property $ do
    n <- forAll $ Gen.int64 Range.linearBounded
    test . runResourceT $ do
      -- test with resource usage here
```

does not work with resourece-1.3 though. So if I only add resourcet and hedgehog as dependencies in an application (no upper bounds) I can't make this work. Not sure what's the best/common way to deal with this problem.

PR #222, opened 2018-09-11
Remove the dependency on the unix package

csamak:
I could be missing something big here, but detectMark doesn't seem like it's doing much and it's the only thing using the unix dep.

getEffectiveUserName caused tests to fail on the scratch image I use in my CI setup. I could fix that by populating passwd, but it seemed cleaner to remove it here if it's not needed.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -75,10 +75,6 @@ library
     , transformers-base               >= 0.4        && < 0.5
     , wl-pprint-annotated             >= 0.0        && < 0.2
 
-  if !os(windows)
-    build-depends:
-      unix                            >= 2.6        && < 2.8
-
   ghc-options:
     -Wall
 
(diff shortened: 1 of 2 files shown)


tonymorris:
How else might we detect Mark?

thumphries:
Going to put @jystic on the spot for this one :sweat_smile:

thumphries:
Sorry for the slow follow-up.

We're going to keep this easter egg here by popular demand, but we should catch the exception so it causes no problems for users in your situation.

tonymorris:
I think there should be a `hedgehog-mark` package with a `unix` and `hedgehog` dependency, while `hedgehog` does not have the dependency.

This way the importance of detecting Mark is preserved while the orthogonal practical consequences (such as a `unix` dependency and consequently, eta) can be improved.

thumphries:
Solved another way in #226 , thanks all

PR #224, opened 2018-09-25
Update upper bounds for many dependencies

gwils:
Leave resourcet in hedgehog-example for now, since
it does not compile with newer resourcet

With these bounds in place, we should be able to build on GHC 8.6.1 once lifted-async has updated their base upper bound. When that happens I'll add 8.6.1 to the travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -126,7 +126,7 @@ script:
 
   # build & run tests, build benchmarks
   - cabal new-build -w ${HC} ${TEST} ${BENCH} all
-  - if [ "x$TEST" = "x--enable-tests" ]; then cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
+  - if [ "x$TEST" = "x--enable-tests" ]; then travis_wait 30 cabal new-test -w ${HC} ${TEST} ${BENCH} all; fi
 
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -28,6 +28,9 @@ before_cache:
 
 matrix:
   include:
+    - compiler: "ghc-8.6.1"
+    # env: TEST=--disable-tests BENCH=--disable-benchmarks
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.6.1], sources: [hvr-ghc]}}
```
moodmosaic: Shouldn't we want `cabal-install-2.4` instead?

.travis.yml
```diff
@@ -2,7 +2,7 @@
 #
 #   runghc make_travis_yml_2.hs 'cabal.project'
 #
-# For more information, see https://github.com/hvr/multi-ghc-travis
+# For more information, see https://github.com/haskell-CI/haskell-ci
```
moodmosaic: :+1:

.travis.yml
gwils: Excellent catch

PR #229, opened 2018-10-26
Expose discoverPrefix to find prefixed properties

ruhatch:
We are using this change to have specific sets of properties, for example golden tests, with their own prefix.


jystic:
This is a nice change, thanks for the PR.

PR #233, opened 2018-11-15
Add MonadFail instances where appropriate

HuwCampbell:
Some folks might use generators in do notation with a partial pattern match, this won't work with GHC 8.6 without a MonadFail instance.

Although it's probably not the best practice, I would imagine that discarding the non-matching sample is better than throwing an exception.

~~Other that that I just turned on a few warnings and fixed them.~~

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -58,6 +58,7 @@ library
     , containers                      >= 0.4        && < 0.7
     , directory                       >= 1.2        && < 1.4
     , exceptions                      >= 0.7        && < 0.11
+    , fail                            == 4.9.*
     , lifted-async                    >= 0.7        && < 0.11
     , mmorph                          >= 1.0        && < 1.2
     , monad-control                   >= 1.0        && < 1.1
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -58,7 +58,7 @@ library
     , containers                      >= 0.4        && < 0.7
     , directory                       >= 1.2        && < 1.4
     , exceptions                      >= 0.7        && < 0.11
-    , fail                            == 4.9.*
+    , fail                            >= 4.9        && < 5
     , lifted-async                    >= 0.7        && < 0.11
     , mmorph                          >= 1.0        && < 1.2
     , monad-control                   >= 1.0        && < 1.1
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -158,6 +158,8 @@ import           Control.Monad (MonadPlus(..), filterM, replicateM, ap, join)
 import           Control.Monad.Base (MonadBase(..))
 import           Control.Monad.Catch (MonadThrow(..), MonadCatch(..))
 import           Control.Monad.Error.Class (MonadError(..))
+import           Control.Monad.Fail (MonadFail (..))
+import qualified Control.Monad.Fail as Fail
 import           Control.Monad.IO.Class (MonadIO(..))
 import           Control.Monad.Morph (MFunctor(..), MMonad(..), generalize)
 import           Control.Monad.Primitive (PrimMonad(..))
@@ -552,6 +554,13 @@ instance Monad m => Monad (GenT m) where
         (sk, sm) ->
           runGenT size sk . k =<<
           runGenT size sm m
+
+  fail =
+    Fail.fail
+
+instance Monad m => MonadFail (GenT m) where
+  fail =
+    error
 
 instance Monad m => Alternative (GenT m) where
   empty =
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -84,6 +84,8 @@ import           Control.Monad.Base (MonadBase(..))
 import           Control.Monad.Catch (MonadThrow(..), MonadCatch(..))
 import           Control.Monad.Catch (SomeException(..), displayException)
 import           Control.Monad.Error.Class (MonadError(..))
+import           Control.Monad.Fail (MonadFail (..))
+import qualified Control.Monad.Fail as Fail
 import           Control.Monad.IO.Class (MonadIO(..))
 import           Control.Monad.Morph (MFunctor(..))
 import           Control.Monad.Primitive (PrimMonad(..))
@@ -317,6 +319,10 @@ instance Monad m => Monad (TestT m) where
       unTest m >>=
       unTest . k
 
+  fail =
+    Fail.fail
+
+instance Monad m => MonadFail (TestT m) where
   fail err =
     TestT . ExceptT . pure . Left $ Failure Nothing err Nothing
 


hedgehog/hedgehog.cabal
```diff
@@ -76,8 +76,11 @@ library
     , transformers-base               >= 0.4        && < 0.5
     , wl-pprint-annotated             >= 0.0        && < 0.2
 
+  if !impl(ghc >= 8)
+    build-depends:     fail           == 4.9.*
+
   ghc-options:
-    -Wall
+    -Wall -Wcompat -Wnoncanonical-monad-instances -Wnoncanonical-monadfail-instances
```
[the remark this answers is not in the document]
HuwCampbell: I think that's reasonable.

hedgehog/hedgehog.cabal
```diff
@@ -76,8 +76,11 @@ library
     , transformers-base               >= 0.4        && < 0.5
     , wl-pprint-annotated             >= 0.0        && < 0.2
 
+  if !impl(ghc >= 8)
+    build-depends:     fail           == 4.9.*
```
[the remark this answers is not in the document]
HuwCampbell: Ok

HuwCampbell:
If we only provide a MonadFail instance for GenT (as in you diff above) without a `fail = Fail.fail` implementation in Monad, one will see very different behaviours depending on whether -XMonadFailDesugaring is on or off. Wouldn't be ideal.

HuwCampbell:
I've minimised this PR to the smallest workable solution, will do the canonical versions in another PR.

moodmosaic:
> If we only provide a MonadFail instance for GenT (as in you diff above) without a `fail = Fail.fail`

There seems to be a pending discussion around [this](https://prime.haskell.org/wiki/Libraries/Proposals/MonadFail#Adaptingoldcode) with plans to remove `fail` in GHC 8.8+

>one will see very different behaviours depending on whether -XMonadFailDesugaring is on or off. 

I didn't see any difference between GHC 7.10.3, 8.2.2, and 8.6.2 after taking out `fail = Fail.fail`:

```haskell
maybeOne :: Gen (Maybe Int)
maybeOne = do
  Just n <-
    pure <$> Gen.constant (Just 1)
  return n

prop_maybeOne :: Property
prop_maybeOne =
  property $ do
    Just x <- forAll maybeOne
    assert $
      x == 1
```

That's why I didn't include `fail = Fail.fail` in the diff. Do you have an example?

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -317,6 +319,10 @@ instance Monad m => Monad (TestT m) where
       unTest m >>=
       unTest . k
 
+  fail =
+    Fail.fail
+
+instance Monad m => MonadFail (TestT m) where
   fail err =
```
moodmosaic: We need one for `PropertyT` as well

```haskell
instance MonadPlus m => MonadFail (PropertyT m) where
  fail _ =
    mzero
```

---

Here's an example:

```haskell
maybeOne :: Gen (Maybe Int)
maybeOne = do
  Just n <-
    pure <$> Gen.constant (Just 1)
  return n

prop_maybeOne :: Property
prop_maybeOne =
  property $ do
    Just x <- forAll maybeOne
    assert $
      x == 1
```

This doesn't compile without a `MonadFail (PropertyT m)` instance:

```
        * No instance for (Control.Monad.Fail.MonadFail (PropertyT IO))
            arising from a do statement
            with the failable pattern `Just x'
        * In a stmt of a 'do' block: Just x <- forAll maybeOne
          In the second argument of `($)', namely
            `do Just x <- forAll maybeOne
                assert $ x == 1'
          In the expression:
            property
              $ do Just x <- forAll maybeOne
                   assert $ x == 1
       |
    39 |     Just x <- forAll maybeOne
       |     ^^^^^^^^^^^^^^^^^^^^^^^^^
```

HuwCampbell:
That example never hits the Nothing case in the pattern match and exercises MonadFail.

A simple change is from the hedgehog project in `module Test.Hedgehog.Text`. This won't fail on exception (but will fail on discard half the time) with this change.

```haskell
genOdd :: Gen Int64
genOdd =
  let
    isOdd x =
      if odd x then
        Just x
      else
        Nothing
  in do
    Just x <- isOdd <$> Gen.int64 (Range.constant 1 maxBound)
    return x
```

hedgehog/src/Hedgehog/Internal/Property.hs
HuwCampbell: I considered that as well but thought the semantics wouldn't be very clear. For a generator, `mzero` and `fail` being discard is pretty clear. While for a test, `fail` being a concrete failure is pretty clear. For a property... I don't think there's a single correct answer, so I omitted the instance.

moodmosaic: Yeah, that's true. Better be explicit and let the compiler do its job, then.

moodmosaic:
> That example never hits the Nothing case in the pattern match and exercise MonadFail.

Indeed, good catch!

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -553,6 +555,13 @@ instance Monad m => Monad (GenT m) where
           runGenT size sk . k =<<
           runGenT size sm m
 
+  fail =
+    Fail.fail
+
+instance Monad m => MonadFail (GenT m) where
+  fail _ =
+    mzero
```
moodmosaic: See https://github.com/hedgehogqa/haskell-hedgehog/pull/233#pullrequestreview-177737024.

hedgehog/src/Hedgehog/Internal/Property.hs
moodmosaic: However, now that we've agreed to go with `fail = error` until the next major version bump, it might be worth considering

```haskell
instance MonadPlus m => MonadFail (PropertyT m) where
  fail =
    error
```

here as well. ━ Just we don't break any builds for those switching to GHC 8.6.

HuwCampbell: People expect to see MonadFail errors when upgrading to 8.6. It's part and parcel with the upgrade and that's a good thing. We shouldn't add a `MonadFail` instances just to counter that.

moodmosaic: Yeah, I think `fail` for a generator and a test is fair enough.

hedgehog/hedgehog.cabal
```diff
@@ -58,6 +58,7 @@ library
     , containers                      >= 0.4        && < 0.7
     , directory                       >= 1.2        && < 1.4
     , exceptions                      >= 0.7        && < 0.11
+    , fail                            == 4.9.*
```
jystic: Would you mind keeping the same style as the other version number numbers? Makes it easier to bump the upper bounds in the future. Other than that I'm happy with these changes because I can do a release without bumping a major version. We can talk about the tradeoffs of `mzero` in another PR perhaps, but in general I don't like breaking changes if I can avoid it.

HuwCampbell: Ok

moodmosaic: @jystic, @HuwCampbell just [forced-pushed](https://github.com/hedgehogqa/haskell-hedgehog/compare/d770d507dc7ab9007a9028786a11e569dbb52049..b400bf14216161e67048f4adbcd71e45bde5e0f6) the change so it's looking good now :+1:

hedgehog/src/Hedgehog/Internal/Property.hs
moodmosaic: /cc @fosskers

PR #235, opened 2018-11-20
Show output variables when parallel state machine testing fails to linearise

HuwCampbell:
`no valid interleaving` as an error message leaves a bit to be desired. Without seeing what the concrete outputs were, it can be hard to figure out where the race condition(s) might be.

With this change, we show the outputs each Concrete Var got during the parallel execution.

The biggest downside is that all output types must now have `Show` instance, so it's unfortunately a breaking change.

screen shot 2018-11-20 at 12 19 44 pm


jystic:
This is cool, I wish my output system was more flexible so we could merge the `Gen` and the `executeParallel` output, but I really like this nonetheless.

Issue #72, opened 2017-05-06
Hackage badge is broken


jystic:
The shields.io [server](https://github.com/badges/shields/blob/6031711180f1576f78ce654b693a55ab1540468d/server.js#L2778-L2781) doesn't parse .cabal files in a particularly robust way, so it can't figure out the version number of Hedgehog.

I'd like to keep the formatting that we have for the `hedgehog.cabal` file, so it would be good to fix the shields.io cabal file parser at some point. Essentially is should be changed to find the word following `/^version:/`, rather than what it does currently, which is assume that the whole thing is on one line.

HuwCampbell:
Dancing around the problem: one can fudge a hackage badge with a version number manually with
```
[hackage-png]: https://img.shields.io/badge/hackage-v0.2.2.0-orange.svg
```

jystic:
> Dancing around the problem: one can fudge a hackage badge with a version number manually with

This is what i've done for now, would still like to get it fixed upstream though.

PR #239, opened 2018-12-09
Bump Badge

HuwCampbell:


README.md
```diff
@@ -77,7 +77,7 @@ You can then load the module in GHCi, and run it:
 ```
 
  [hackage]: http://hackage.haskell.org/package/hedgehog
- [hackage-shield]: https://img.shields.io/badge/hackage-v0.5-blue.svg
+ [hackage-shield]: https://img.shields.io/badge/hackage-v0.6.1-blue.svg
```
moodmosaic: Is there a way to make this dynamic so we don't forget to update it every time there's a version bump?

HuwCampbell: Unfortunately their parser is broken for cabal files and requires the version to be written on the same line as `version:`. See #72

Issue #201, opened 2018-05-27
more convenient way to use a fixed seed for a test group


ghorn:
I would like to use our extensive hedgehog test suite on my continuous integration server, but fortunately and unfortunately it occasionally finds new corner cases and fails commits which have nothing to do with the hedgehog tests so I had to disable them.

The solution seems to be to always run with the same seed on CI to prevent regressions but not find new issues, and then some mega nightly build with a random seed to look for new failures.

The problem is the only seeded test function I could find is `recheck` which serves a different purpose. Is there a reason that `Hedgehog.Internal.Runner.checkGroupWith` can't take a seed? I could jigger `recheck` to do this but I would be missing out on all the advantages of using `checkGroup`.

ghorn:
If someone could provide a few pointers, I'd be happy to do this work.

thumphries:
Hey! Sorry for the slow.

We should be able to support this workflow with a couple of changes. If you float a `Maybe Seed` up into the signature for `checkGroupWith`, that is probably a good start. We can bikeshed the API change once it's working.

Currently each property gets a new seed in [`checkNamed`](https://github.com/hedgehogqa/haskell-hedgehog/blob/a506b6d2ec9b8721aead61b72a2f1fa8b4c88ce5/hedgehog/src/Hedgehog/Internal/Runner.hs#L238), which calls `checkRegion`, which takes a `Seed`. I'm not too sure what the behaviour should be with a fixed seed, but you might want to split it before each property.

It's a little outside the way most people use Hedgehog, usually the continuous search for bugs in CI is a feature. But I see where you're coming from and we should have a better story. Anecdotally I have heard of teams discouraged from doing property-based testing in monorepo situations, which is pretty terrible!

thumphries:
An alternative design: read `HEDGEHOG_SEED` from `ENV` and use that if defined. Unsure if wise

ghorn:
Thanks for the pointers. I strongly prefer the API change over the environment variable. Let me give it a try.

moodmosaic:
Here's a patch for setting a fixed seed through the `HEDGEHOG_SEED` environment variable, which has to be in the form of `15417483394291888538 15013407834460730181`.

```diff
diff --git a/hedgehog/src/Hedgehog/Internal/Config.hs b/hedgehog/src/Hedgehog/Internal/Config.hs
index c679959..efd3320 100644
--- a/hedgehog/src/Hedgehog/Internal/Config.hs
[...]

PR #240, opened 2018-12-12
Fix minor performance issue with color handling

christianlavoie:
Turns out `hSupportsANSI` is not the most efficient call:

```
$ cat spec-before.prof | head -20
	Tue Dec 11 20:42 2018 Time and Allocation Profiling Report  (Final)

	   spec +RTS -p -s -t -RTS

	total time  =        2.03 secs   (2026 ticks @ 1000 us, 1 processor)
	total alloc = 3,087,822,320 bytes  (excludes profiling overheads)

COST CENTRE             MODULE                   SRC                                              %time %alloc

>>=                     Hedgehog.Internal.Tree   src/Hedgehog/Internal/Tree.hs:(136,3)-(141,29)    11.4   11.5
hSupportsANSI.isNotDumb System.Console.ANSI.Unix src/System/Console/ANSI/Unix.hs:74:3-65            8.3   20.6
<*>                     Hedgehog.Internal.Gen    src/Hedgehog/Internal/Gen.hs:(542,3)-(543,6)       7.1    7.1
split                   Hedgehog.Internal.Seed   src/Hedgehog/Internal/Seed.hs:(135,1)-(140,39)     4.3    6.0
randomIvalInteger       System.Random            System/Random.hs:(468,1)-(489,76)                  4.0    3.2
runGenT                 Hedgehog.Internal.Gen    src/Hedgehog/Internal/Gen.hs:(233,1)-(234,13)      4.0    0.2
next                    Hedgehog.Internal.Seed   src/Hedgehog/Internal/Seed.hs:(126,1)-(130,17)     3.3    7.0
list.\                  Hedgehog.Internal.Gen    src/Hedgehog/Internal/Gen.hs:(1302,5)-(1306,31)    3.3    2.6
fmap                    Hedgehog.Internal.Tree   src/Hedgehog/Internal/Tree.hs:(107,3)-(108,34)     2.4    2.9
expand                  Hedgehog.Internal.Tree   src/Hedgehog/Internal/Tree.hs:(85,1)-(89,44)       2.2    2.2
>>=.\                   Hedgehog.Internal.Gen    src/Hedgehog/Internal/Gen.hs:(551,7)-(554,27)      2.0    4.0
```

This is somewhat of a problem because as it stands, hedgehog aggressively checks if the same terminal supports color every property run. This PR makes it check ANSI support once before running and never after, resulting in a 10ish% CPU saving on a CPU-intensive (if quite short) test:

```
$ cat spec-after.prof | head -20
	Tue Dec 11 21:07 2018 Time and Allocation Profiling Report  (Final)

	   spec +RTS -p -s -t -RTS

	total time  =        2.03 secs   (2030 ticks @ 1000 us, 1 processor)
	total alloc = 2,275,820,544 bytes  (excludes profiling overheads)

COST CENTRE            MODULE                     SRC                                                 %time %alloc

>>=                    Hedgehog.Internal.Tree     src/Hedgehog/Internal/Tree.hs:(136,3)-(141,29)       11.2   14.5
<*>                    Hedgehog.Internal.Gen      src/Hedgehog/Internal/Gen.hs:(542,3)-(543,6)          7.8    9.0
split                  Hedgehog.Internal.Seed     src/Hedgehog/Internal/Seed.hs:(135,1)-(140,39)        5.4    7.6
randomIvalInteger      System.Random              System/Random.hs:(468,1)-(489,76)                     4.3    4.1
runGenT                Hedgehog.Internal.Gen      src/Hedgehog/Internal/Gen.hs:(233,1)-(234,13)         3.7    0.3
next                   Hedgehog.Internal.Seed     src/Hedgehog/Internal/Seed.hs:(126,1)-(130,17)        3.7    8.9
list.\                 Hedgehog.Internal.Gen      src/Hedgehog/Internal/Gen.hs:(1302,5)-(1306,31)       3.3    3.2
>>=.\                  Hedgehog.Internal.Gen      src/Hedgehog/Internal/Gen.hs:(551,7)-(554,27)         3.0    5.1
expand                 Hedgehog.Internal.Tree     src/Hedgehog/Internal/Tree.hs:(85,1)-(89,44)          2.9    2.8
fmap                   Hedgehog.Internal.Tree     src/Hedgehog/Internal/Tree.hs:(107,3)-(108,34)        2.5    3.7
integral_.\            Hedgehog.Internal.Gen      src/Hedgehog/Internal/Gen.hs:(804,5)-(809,57)         2.4    2.3
```

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -78,7 +78,6 @@ library
 
   ghc-options:
     -Wall
-    -Werror
 
   hs-source-dirs:
     src
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -78,6 +78,7 @@ library
 
   ghc-options:
     -Wall
+    -Werror
 
   hs-source-dirs:
     src
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -784,7 +784,7 @@ ppSummary summary =
             Nothing
         ]
 
-renderDoc :: MonadIO m => Maybe UseColor -> Doc Markup -> m String
+renderDoc :: MonadIO m => UseColor -> Doc Markup -> m String
 renderDoc mcolor doc = do
   let
     dull =
@@ -879,11 +879,9 @@ renderDoc mcolor doc = do
     end _ =
       setSGRCode [Reset]
 
-  color <- resolveColor mcolor
-
   let
     display =
-      case color of
+      case mcolor of
         EnableColor ->
           WL.displayDecorated start end id
         DisableColor ->
@@ -899,14 +897,14 @@ renderDoc mcolor doc = do
     WL.renderSmart 100 $
     WL.indent 2 doc
 
-renderProgress :: MonadIO m => Maybe UseColor -> Maybe PropertyName -> Report Progress -> m String
+renderProgress :: MonadIO m => UseColor -> Maybe PropertyName -> Report Progress -> m String
 renderProgress mcolor name x =
   renderDoc mcolor =<< ppProgress name x
 
-renderResult :: MonadIO m => Maybe UseColor -> Maybe PropertyName -> Report Result -> m String
+renderResult :: MonadIO m => UseColor -> Maybe PropertyName -> Report Result -> m String
 renderResult mcolor name x =
   renderDoc mcolor =<< ppResult name x
 
-renderSummary :: MonadIO m => Maybe UseColor -> Summary -> m String
+renderSummary :: MonadIO m => UseColor -> Summary -> m String
 renderSummary mcolor x =
   renderDoc mcolor =<< ppSummary x
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -240,7 +240,9 @@ recheck size seed prop0 = do
     checkRegion region mcolor Nothing size seed prop
   pure ()
 
-data Parallelism = UseParallelism | NoParallelism
+data Parallelism =
+    UseParallelism
+  | NoParallelism
 
 -- | Check a group of properties using the specified runner config.
 --
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -56,15 +56,15 @@ data RunnerConfig =
   RunnerConfig {
       -- | The number of property tests to run concurrently. 'Nothing' means
       --   use one worker per processor.
-      runnerWorkers :: !(Maybe WorkerCount)
+      runnerWorkers :: !WorkerCount
 
       -- | Whether to use colored output or not. 'Nothing' means detect from
       --   the environment.
-    , runnerColor :: !(Maybe UseColor)
+    , runnerColor :: !UseColor
 
       -- | How verbose to be in the runner output. 'Nothing' means detect from
       --   the environment.
-    , runnerVerbosity :: !(Maybe Verbosity)
+    , runnerVerbosity :: !Verbosity
     } deriving (Eq, Ord, Show)
 
 findM :: Monad m => [a] -> b -> (a -> m (Maybe b)) -> m b
@@ -203,7 +203,7 @@ checkReport cfg size0 seed0 test0 updateUI =
 checkRegion ::
      MonadIO m
   => Region
-  -> Maybe UseColor
+  -> UseColor
   -> Maybe PropertyName
   -> Size
   -> Seed
@@ -234,7 +234,7 @@ checkRegion region mcolor name size seed prop =
 checkNamed ::
      MonadIO m
   => Region
-  -> Maybe UseColor
+  -> UseColor
   -> Maybe PropertyName
   -> Property
   -> m (Report Result)
@@ -244,26 +244,26 @@ checkNamed region mcolor name prop = do
 
 -- | Check a property.
 --
-check :: MonadIO m => Property -> m Bool
-check prop =
+check :: MonadIO m => UseColor -> Property -> m Bool
+check mcolor prop =
   liftIO . displayRegion $ \region ->
-    (== OK) . reportStatus <$> checkNamed region Nothing Nothing prop
+    (== OK) . reportStatus <$> checkNamed region mcolor Nothing prop
 
 -- | Check a property using a specific size and seed.
 --
-recheck :: MonadIO m => Size -> Seed -> Property -> m ()
-recheck size seed prop0 = do
+recheck :: MonadIO m => UseColor -> Size -> Seed -> Property -> m ()
+recheck mcolor size seed prop0 = do
   let prop = withTests 1 prop0
   _ <- liftIO . displayRegion $ \region ->
-    checkRegion region Nothing Nothing size seed prop
+    checkRegion region mcolor Nothing size seed prop
   pure ()
 
 -- | Check a group of properties using the specified runner config.
 --
-checkGroup :: MonadIO m => RunnerConfig -> Group -> m Bool
-checkGroup config (Group group props) =
+checkGroup :: MonadIO m => Bool -> Group -> m Bool
+checkGroup useParallelism (Group group props) =
   liftIO $ do
-    n <- resolveWorkers (runnerWorkers config)
+    n <- if useParallelism then detectWorkers else pure 1
 
     -- ensure few spare capabilities for concurrent-output, it's likely that
     -- our tests will saturate all the capabilities they're given.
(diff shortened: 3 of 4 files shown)


hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,27 @@ checkNamed region mcolor name prop = do
...
   _ <- liftIO . displayRegion $ \region ->
-    checkRegion region Nothing Nothing size seed prop
+    checkRegion region mcolor Nothing size seed prop
   pure ()
 
 -- | Check a group of properties using the specified runner config.
 --
-checkGroup :: MonadIO m => RunnerConfig -> Group -> m Bool
-checkGroup config (Group group props) =
+checkGroup :: MonadIO m => Bool -> Group -> m Bool
```
thumphries: Could we get something like `data Parallelism = UseParallelism | NoParallelism` here instead of the Bool?

hedgehog/hedgehog.cabal
```diff
@@ -78,6 +78,7 @@ library
 
   ghc-options:
     -Wall
+    -Werror
```
thumphries: `-Werror` is discouraged in public libraries, since warnings could be introduced by newer versions of GHC, making things a bit too brittle for downstream (I believe `cabal lint` complains about it)

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -784,7 +784,7 @@ ppSummary summary =
             Nothing
         ]
 
-renderDoc :: MonadIO m => Maybe UseColor -> Doc Markup -> m String
+renderDoc :: MonadIO m => UseColor -> Doc Markup -> m String
 renderDoc mcolor doc = do
```
thumphries: This variable name is short for `maybe color`, if it's not a `Maybe` anymore we should rename it

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,31 @@ checkNamed region mcolor name prop = do
 -- | Check a property.
 --
 check :: MonadIO m => Property -> m Bool
-check prop =
+check prop = do
+  mcolor <- detectColor
```
thumphries: same here

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,31 @@ checkNamed region mcolor name prop = do
...
+
 -- | Check a group of properties using the specified runner config.
 --
-checkGroup :: MonadIO m => RunnerConfig -> Group -> m Bool
-checkGroup config (Group group props) =
+checkGroup :: MonadIO m => Parallelism -> Group -> m Bool
+checkGroup parallelism (Group group props) =
   liftIO $ do
-    n <- resolveWorkers (runnerWorkers config)
+    n <- case parallelism of
```
thumphries: could we stick to the predominant style here please - new lines for each case

```haskell
case parallelism of
  UseParallelism ->
    detectWorkers
  NoParallelism ->
    pure 1
```

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,31 @@ checkNamed region mcolor name prop = do
...
+  mcolor <- detectColor
   liftIO . displayRegion $ \region ->
-    (== OK) . reportStatus <$> checkNamed region Nothing Nothing prop
+    (== OK) . reportStatus <$> checkNamed region mcolor Nothing prop
 
 -- | Check a property using a specific size and seed.
 --
 recheck :: MonadIO m => Size -> Seed -> Property -> m ()
 recheck size seed prop0 = do
+  mcolor <- detectColor
```
thumphries: and here

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,31 @@ checkNamed region mcolor name prop = do
...
   let prop = withTests 1 prop0
   _ <- liftIO . displayRegion $ \region ->
-    checkRegion region Nothing Nothing size seed prop
+    checkRegion region mcolor Nothing size seed prop
   pure ()
 
+data Parallelism = UseParallelism | NoParallelism
```
thumphries: style here should be consistent

```haskell
data Parallelism =
    UseParallelism
  | NoParallelism
```

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -245,25 +225,33 @@ checkNamed region mcolor name prop = do
...
 
+data Parallelism =
+    UseParallelism
+  | NoParallelism
+
 -- | Check a group of properties using the specified runner config.
 --
-checkGroup :: MonadIO m => RunnerConfig -> Group -> m Bool
-checkGroup config (Group group props) =
+checkGroup :: MonadIO m => Parallelism -> Group -> m Bool
```
moodmosaic: What about all the other options we had when passing a `RunnerConfig`? We were able to specify color, verbosity, workers, (even a fixed seed as discussed in https://github.com/hedgehogqa/haskell-hedgehog/issues/201#issuecomment-441401915).

Could you please explain the motivation behind this breaking change?

christianlavoie: Fixed seed wasn't part of the `RunnerConfig` I removed -- only verbosity, color and workers.

The first version I wrote (privately, never pushed here) actually did just removed the `Maybe` part of the `RunnerConfig` and forced choice early on all three fields... until I noticed that only color was every checked in a tight loop, the other weren't. Workers doesn't make sense to check after starting to run properties anyway, and verbosity seemed (at a quick glance and lack of compiler complaints from removing it altogether) like an artifact from before progress formatters were implemented. I didn't check in depth, but it doesn't actually seem to do anything at all.

In the end, I noticed neither workers nor (it seemed) verbosity were used after running properties had started, so I just gutted everything to make that fact more explicit. The change doesn't actually break anything, terminals can't change color support halfway through (unless you really enjoy changing your term's settings while hedgehog is running) and the other two weren't actually changing anything after property checking started.

It could make sense to keep `RunnerConfig` around, I don't feel strongly. I just guessed it was an artifact from the Ye Olde Days based on how it was (not) used. Lemme know.

moodmosaic: :+1: Having only the minimum required changes in this PR will increase the chances of getting it merged sooner, at least until @jystic is back. We can always tweak/remove `RunnerConfig` separately.

jystic:
I like this change in general but I'm finding it a bit hard to review with the `RunnerConfig` removal as well. If you could keep `RunnerConfig` for now and maybe propose removing it in a separate PR I'd be happy to apply the performance change. That way we can all discuss its merits as it really is a different issue.

stolyaroleh:
I'd like to point out that the performance hit is not that minor. I just stumbled upon this while writing a simple test for binary search. Given that every iteration was doing <50 `Int` comparisons, it was quite surprising to discover that the `-O2` compiled version took more than few milliseconds to run 1000 iterations!

Here's some more numbers I collected.
With `hSupportsANSI`:
```
	total time  =        0.44 secs   (1326 ticks @ 1000 us, 3 processors)
	total alloc = 2,418,786,712 bytes  (excludes profiling overheads)

COST CENTRE           MODULE                             SRC                                                     %time %alloc

hSupportsANSI         System.Console.ANSI.Unix           src/System/Console/ANSI/Unix.hs:(71,1)-(74,65)           71.5   90.3
async                 Control.Concurrent.Async           Control/Concurrent/Async.hs:194:1-35                     12.7    4.9
```
Without:
```
	total time  =        0.05 secs   (153 ticks @ 1000 us, 3 processors)
	total alloc = 101,967,872 bytes  (excludes profiling overheads)

COST CENTRE             MODULE                        SRC                                                      %time %alloc

async                   Control.Concurrent.Async      Control/Concurrent/Async.hs:194:1-35                      47.1   36.8
runGenT                 Hedgehog.Internal.Gen         src/Hedgehog/Internal/Gen.hs:(240,1)-(241,13)              8.5    7.2
evalIO                  Hedgehog.Internal.Property    src/Hedgehog/Internal/Property.hs:(755,1)-(756,71)         7.8    7.4
```
1000 iterations of a simple test allocate 2.4 GB!

I'd like to bump this PR. If @christianlavoie does not mind, I can implement changes @moodmosaic requested either on top of this PR or as a separate one, with just one small commit.

moodmosaic:
@stolyaroleh, sure, go ahead :+1:

IIRC, the most important thing was to keep the existing stuff in `RunnerConfig`, and so then we can later decide to clean up separately if and when needed, as per https://github.com/hedgehogqa/haskell-hedgehog/pull/240#issuecomment-471242309.

moodmosaic:
Thanks for all the work @christianlavoie and @stolyaroleh. :+1: :rocket: Closing this as it's resolved via #345. Feel free to reopen or add comments to it, if needed.

PR #247, opened 2019-02-14
Bump ansi-terminal upper bound

gwils:
I've already made the metadata revision corresponding to this pull request, so there's no need for a new release.


thumphries:
Thanks George!

Issue #122, opened 2017-10-19
We should document that PropertyT is like a non-deterministic monad


ocharles:
I was just playing around with interleaving some database calls in a `PropertyT` computation, and confused myself. I've now come to realise that `PropertyT` is like a non-deterministic monad (e.g., list, `LogicT`) and that performing IO doesn't necessarily imply it will all be performed in exactly the same order over shrinks. For example,

```haskell
property $ do
  foo <- forAll $ pure True
  liftIO $ putStrLn "blah"
  bar <- forAll $ Gen.integral (Range.linear 1 100)
  ...
```

Will only ever execute the `putStrLn` once, as everything "above" it is a one-node `GenT` tree (as confirmed by `pure` with `forAll`). If something fails on the value of `bar`, the next thing to do is shrink `bar`, and that is where execution will resume from.

If I'm correct in my understanding, I think it would be good to write this down.

jystic:
Yeah that's correct, it's always been a concern of mine that allowing effects like this is a bit confusing, but I didn't want to rule it out in case someone found a good use case for it.

jystic:
I think the example in the above post could be made to work using `ReaderT`, something like:

```haskell
s_coin :: (
    Monad gen
  , MonadIO m
  , MonadTest m
  , MonadReader Turnstile m
  )
  => Turnstile
  -> Command gen m ModelState
s_coin =
  ...

prop_turnstile :: Property
prop_turnstile =
  property $ do
    actions <- forAll $
      Gen.sequential (Range.linear 1 100) initialState [
          s_coin
        , s_push_locked
        , s_push_unlocked
        ]
    turnstile <- liftIO newTurnstile
    flip runReaderT turnstile $ executeSequential initialState actions
```

massysett:
Hi Jacob,

Thanks for all your work on Hedgehog.  It's a huge advance in the realm of property testing, and it has saved me a lot of work and frustration.  The elegance of the library and the thought that went into its API is impressive.

[...]

PR #253, opened 2019-03-16
Add classifications to properties

felixmulder:
Was recommended this great talk by John Hughes on labelling properties, thought that would be cool to have in Hedgehog too.


hedgehog/hedgehog.cabal
```diff
@@ -75,6 +75,7 @@ library
     , time                            >= 1.4        && < 1.10
     , transformers                    >= 0.4        && < 0.6
     , transformers-base               >= 0.4        && < 0.5
+    , unordered-containers
```
felixmulder: Not sure what version to set here yet - I guess travis will show us what works for all GHCs?

felixmulder: Resolved this in latest fpush

felixmulder:
Tried a bunch of things to work around the bug in GHC 8.2.1. Couldn't get it unborken tonight. How important is 8.2.1 support? 8.2.2 works fine...

moodmosaic:
Thanks for sending a pull request 👍 Consider adding a *[WIP]* on the title if it's not finalized (from what I can tell you keep adding commits). Also, considering squashing all those `wat` commits once you're done.

felixmulder:
Lol, yes sorry - I wanted to fix the 8.2.1 issue before cleaning it up.
I'll add the WIP tag

felixmulder:
@moodmosaic - I ended up figuring out which line was offending 8.2.1. I filter it out on 8.2.x series, maybe there's a more precise way to do this, but I can't figure one out.

So tl;dr, ready for review

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -122,6 +129,9 @@ import           Hedgehog.Internal.Source
 
 import           Language.Haskell.TH.Lift (deriveLift)
 
+import           Prelude hiding (min)
```
moodmosaic: 😕 What if we rename `min` in `cover min True s` to something else instead hiding the built-in one?

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -161,7 +211,7 @@ type Test =
 --
 newtype TestT m a =
   TestT {
-      unTest :: ExceptT Failure (Lazy.WriterT [Log] m) a
+      unTest :: ExceptT Failure (Lazy.WriterT (PropResult, [Log]) m) a
```
moodmosaic: @jystic, perhaps we can take this as-is for now and later refactor this tuple to some data type, e.g. `Journal`. My only concern is that this can easily end up being a triple, quadruple, etc. in the future.

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -111,17 +112,18 @@ takeSmallest ::
...
   Node Nothing _ ->
     pure GaveUp
 
-  Node (Just (x, w)) xs ->
+  Node (Just (x, (_, w))) xs ->
     case x of
       Left (Failure loc err mdiff) -> do
         let
           failure =
+            -- TODO: maybe add classifier to failure?
```
moodmosaic: We use `FIXME` throughout the code base to mark something to be done. See similar nitpick above (shouldn't be a blocker, of course).

felixmulder:
@moodmosaic - I think I addressed most of your comments.

If this patch is accepted, I'd like to continue work on the feature. Things I'd like to do:

- Allow turning coverage warning into failure
- Refactor tuple to data (Journal was a good name 👍)

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -778,6 +831,38 @@ withRetries :: ShrinkRetries -> Property -> Property
...
+                 . unTest . unPropertyT
+  where
+    addClassification m = do
+      (r, (cl@(Classification cls _), xs)) <- m
+#if __GLASGOW_HASKELL__ != 802
+      -- FIXME GHC 8.2.1 has a bug with referencing `cls` below, thus we need to
+      -- FIXME remove these lines if we're using the 802 series
+      when (Map.member s cls) $ throwError . userError $
+        "classification matched duplicate label: \"" <> s <> "\""
+#endif
```
jystic: Could you rewrite this without the pattern match to avoid the CPP?

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -152,6 +161,50 @@ newtype PropertyT m a =
...
+
+-- | A classifier can be attached to a property conditionally
+--
+--   When the amount of occurrences don't exceed the minimum percentage, a
+--   warning will be issued.
+data Classifier = Classifier
+  { clsMinPercentage :: !Double
+  , clsOccurrences :: !Integer
+  }
+  deriving Show
```
jystic: I apologise for not having a style guide, but no abbreviations / acronyms where possible (although `x`, `x0`, `x1`, `y` is fine for locals where types are obvious) and also it should be laid out like this:

```haskell
data Classifier =
  Classifier {
      classifierMinimumPercentage :: !Double -- or just classifierPercentage if you prefer
    , classifierOccurrences :: !Integer
    } deriving (Show)
```

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -152,6 +161,50 @@ newtype PropertyT m a =
     , MonadError e
     )
 
+-- | Classifiers use strings as labels
+--
+type ClassifierName = String
```
jystic: Another style thing, no type synonyms please, make this a `newtype`

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -152,6 +161,50 @@ newtype PropertyT m a =
...
+    Classifier percentage (occ1 + occ2)
+
+-- | Classification are a count of how many times a property has ocurred
+--   during a test run
+--
+data Classification = Classification
+  { propClassifiers :: !(Map ClassifierName Classifier)
+  , propTests :: !Integer
+  }
+  deriving Show
```
jystic: Same as my `Classifier` comment

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -152,6 +161,50 @@ newtype PropertyT m a =
...
+
+instance Semigroup Classification where
+  (Classification c1 t1) <> (Classification c2 t2) =
+    Classification
+      (Map.foldrWithKey (Map.insertWith (<>)) c1 c2)
+      (t1 + t2)
+
+instance Monoid Classification where
+  mappend = (<>)
+  mempty = Classification mempty 1
```
jystic: Please put these on new lines, for example:

```haskell
instance Monoid Classification where
  mappend =
    (<>)
  mempty =
    Classification mempty 1
```

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -778,6 +831,38 @@ withRetries :: ShrinkRetries -> Property -> Property
...
+classify :: Bool -> String -> PropertyT IO () -> PropertyT IO ()
+classify = cover 0
+
+cover :: Double -> Bool -> String -> PropertyT IO () -> PropertyT IO ()
+cover _ False _ = id
+cover minCoverage True s = PropertyT
+                 . TestT
+                 . mapExceptT (Lazy.mapWriterT addClassification)
+                 . unTest . unPropertyT
+  where
```
jystic: No `where` clauses please, can you change this to a `let`

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -778,6 +831,38 @@ withRetries :: ShrinkRetries -> Property -> Property
...
+-- @
+classify :: Bool -> String -> PropertyT IO () -> PropertyT IO ()
+classify = cover 0
+
+cover :: Double -> Bool -> String -> PropertyT IO () -> PropertyT IO ()
+cover _ False _ = id
+cover minCoverage True s = PropertyT
+                 . TestT
+                 . mapExceptT (Lazy.mapWriterT addClassification)
+                 . unTest . unPropertyT
```
jystic: No pattern matching in function definitions please, use an 1if` or a `case` here, also indentation:

```
cover :: Double -> Bool -> String -> PropertyT IO () -> PropertyT IO ()
cover minimumCoverage condition description =
  if condition then
    PropertyT
      . TestT
      . mapExceptT (Lazy.mapWriterT addClassification)
      . unTest
      . unPropertyT
  else
    id
```

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +717,52 @@ ppResult name (Report tests discards result) =
...
+ppClassification (Classification cls total)
+  | Map.null cls = mempty
+  | otherwise = (<>) WL.linebreak $ WL.indent 4 . WL.align . WL.vsep $
+    (\(k, v) -> WL.text $ show (occurrenceRate v total) <> "% " <> k) <$> Map.toList cls
+
+occurrenceRate :: Classifier -> Integer -> Double
+occurrenceRate (Classifier _ occurrences) total =
+  let
+    percentage' :: Double
+    percentage' = fromIntegral occurrences / fromIntegral (total - 2) * 100
```
jystic: No primes please, I usually use decimal numbers instead as they are harder to miss when reading the code. 

`x0` `x1` `x2` vs `x` `x'` `x''`

Although here the prime doesn't seem necessary? I guess you are using it to signify that this is the "next percentage".

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +717,52 @@ ppResult name (Report tests discards result) =
...
-        "."
+        "." <+>
+        ppClassification classes <+>
+        ppCoverage classes
+
+ppClassification :: Classification -> Doc Markup
+ppClassification (Classification cls total)
+  | Map.null cls = mempty
+  | otherwise = (<>) WL.linebreak $ WL.indent 4 . WL.align . WL.vsep $
+    (\(k, v) -> WL.text $ show (occurrenceRate v total) <> "% " <> k) <$> Map.toList cls
```
jystic: Sorry I missed this one in the review, can you change these guards to an `if` expression so we can keep two space indenting:

```haskell
  if Map.null cls then
    mempty
  else
    (<>) WL.linebreak $ WL.indent 4 . WL.align . WL.vsep $
      (\(k, v) -> WL.text $ show (occurrenceRate v total) <> "% " <> k) <$> Map.toList cls

File: hedgehog/src/Hedgehog/Internal/Property.hs (hunk 5 above)
Comment (felixmulder) [reply hedgehog/src/Hedgehog/Internal/Property.hs]: I haven't figured out a good way to rephrase that to be perfectly honest. The bug that arises is:



Which, if I understand it correctly comes from when GHC is trying to figure out how to extract something out of a record. The exact thing that crashes is most likely the `cls` part of: `Map.member s cls`. I'm not sure why the line below the cpp clause works as expected.

Do you have any suggestions?

Or am I misunderstanding you - did you mean just "remove the CPP" stuff?

**Comment by felixmulder** [comment]
I'll amend the request in the morning (Stockholm)! Cheers for reviewing!

File: hedgehog/src/Hedgehog/Internal/Property.hs (hunk 5 above)
Comment (jystic) [reply hedgehog/src/Hedgehog/Internal/Property.hs]: Ah, sorry i thought it was the `<- m` line above, brain fart

hmm that is annoying

Comment (jystic) [reply hedgehog/src/Hedgehog/Internal/Property.hs]: OK, leave the CPP in, maybe just add the text from that image to the FIXME

File: hedgehog/src/Hedgehog/Internal/Property.hs [hunk 14]
```diff
@@ -152,6 +162,55 @@ newtype PropertyT m a =
     , MonadError e
     )
 
+-- | Classifiers use strings as labels
+--
+newtype ClassifierName =
+  ClassifierName {
+      unClassifierName :: String
+    } deriving (Show, Eq, Ord)
```
felixmulder: Indented this like the other newtype below

felixmulder:
@jystic - good morning, that should be all of it. Hope I didn't miss anything 🙏

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +718,54 @@ ppResult name (Report tests discards result) =
...
+    percentage :: Double
+    percentage = fromIntegral occurrences / fromIntegral (total - 2) * 100
+    thousandths :: Integer
+    thousandths = round $ percentage * 10
+  in fromIntegral thousandths / 10
+
+ppCoverage :: Classification -> Doc Markup
+ppCoverage (Classification classifiers total)
+  | null coverageLines = mempty
+  | otherwise =
```
jystic: Sorry, I only pointed out each style issue once I was hoping you would fix the similar ones in the rest of the file.

This should be an `if` same as above

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +718,54 @@ ppResult name (Report tests discards result) =
...
+    WL.linebreak <>
+    WL.linebreak <>
+    WL.text "⚠" <>
+    (WL.indent 3 . WL.align . WL.vsep) coverageLines
+  where
```
jystic: Same here no where clauses

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +718,54 @@ ppResult name (Report tests discards result) =
...
+      (\(ClassifierName k, v) -> WL.text $ show (occurrenceRate v total) <> "% " <> k) <$> Map.toList classifiers
+
+occurrenceRate :: Classifier -> Integer -> Double
+occurrenceRate (Classifier _ occurrences) total =
+  let
+    percentage :: Double
+    percentage = fromIntegral occurrences / fromIntegral (total - 2) * 100
+    thousandths :: Integer
+    thousandths = round $ percentage * 10
+  in fromIntegral thousandths / 10
```
jystic: Sorry, this should be:
```haskell
  let
    percentage :: Double
    percentage =
      fromIntegral occurrences / fromIntegral (total - 2) * 100
    thousandths :: Integer
    thousandths =
      round $ percentage * 10
  in
    fromIntegral thousandths / 10
```

The basic rule is to try and keep everything aligned to a multiple of two spaces, so we always wrap after `=` and `in`

felixmulder:
Ok - cool, I think I fixed all the `where`s and guards now!

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -108,8 +114,10 @@ import qualified Control.Monad.Trans.Writer.Strict as Strict
 
 import qualified Data.Char as Char
 import           Data.Functor.Identity (Identity(..))
+import           Data.Map.Strict (Map)
```
HuwCampbell: Just `Data.Map (Map)` is fine as well. The type between the strict and lazy API is the same.

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +718,58 @@ ppResult name (Report tests discards result) =
...
+occurrenceRate (Classifier _ occurrences) total =
+  let
+    percentage :: Double
+    percentage =
+      fromIntegral occurrences / fromIntegral (total - 2) * 100
+    thousandths :: Integer
+    thousandths =
+      round $ percentage * 10
+  in
+    fromIntegral thousandths / 10
```
HuwCampbell: There's a lot of number fudging here. What's the purpose?

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -715,14 +718,58 @@ ppResult name (Report tests discards result) =
...
+ppCoverage (Classification classifiers total) =
+  let
+    coverageLines = foldMap (uncurry renderCoverage) $ Map.toList classifiers
+    renderCoverage (ClassifierName name) cl@(Classifier minRate _) =
+      if occurrenceRate cl total < minRate then
+        pure . WL.text $
+          "Only " <> show (occurrenceRate cl total) <> "% " <> name <>
+          ", but expected " <> show minRate <> "%"
+      else
+        []
```
HuwCampbell: This isn't correct by any means and will display statistically irrelevant errors.

You should use a beta distribution as the prior of the observed class distribution and see if it's consistent (within a tolerance level) with the desired coverage.

HuwCampbell: As a side note, using a Dirichlet prior would be the right thing to do here in the case of exclusive classifications.

But, as we can classify with overlapping cases, I believe a more "one against all" style is the best way to go.

HuwCampbell: These results from your PR show the problem.

```
  ✓ prop_check_classifiers passed 1 test. 
      51.0% big number
      49.0% small number 
  ✓ prop_check_coverage passed 1 test. 
      51.0% big number
      49.0% small number 
  
  ⚠   Only 49.0% small number, but expected 50.0%
  ✓ 2 succeeded.
```

If we were to turn this into an error (which is kind of the point) we would hit it every time.

hedgehog/src/Hedgehog/Internal/Report.hs
felixmulder: Rounding to 1 decimal place

hedgehog/src/Hedgehog/Internal/Report.hs
felixmulder: This PR does not deal with turning this into a statistically relevant result. If you run X tests it will tell you the result out of those X tests.

I'd like to do what you suggest as a follow-up, because it will allow us to abort the testing if a desired coverage cannot be reached.

hedgehog/src/Hedgehog/Internal/Property.hs
felixmulder: The `Data.Map` suggests using `Data.Map.Strict` when you'll need all values of the map. As such I wanted to be explicit here - but I'd be fine to change it if it's an issue.

HuwCampbell: I'm not suggesting not using the functions from Data.Map.Strict, just noting that the type `Map` is shared, and IMO
```
import           Data.Map (Map)
```
is cleaner

hedgehog/src/Hedgehog/Internal/Report.hs
HuwCampbell: There's also a minus 2?

fenollp:
> Was recommended this great talk by John Hughes on labelling properties, thought that would be cool to have in Hedgehog too.
Sorry do you have a link to the talk?

moodmosaic:
@fenollp

>Sorry do you have a link to the talk?

Yes, https://www.youtube.com/watch?v=NcJOiQlzlXQ

hedgehog/src/Hedgehog/Internal/Report.hs
moodmosaic: I think if we can get rid of those magic numbers—or at least make it more clear for the code reader (remember, code is read more often than it is written)—perhaps we can make some progress on this? /cc @jystic 

On next PRs we can then
* deal with the statistical stuff as commented by @HuwCampbell [here](https://github.com/hedgehogqa/haskell-hedgehog/pull/253/files#r267271586)
* refactor `(Classification, [Log])` into a `Journey` type

felixmulder: So when I run this, it always seems to give an off by two error in the amount of total tests. I can't figure out where it comes from. Hedgehog doesn't seem to be reporting any discarded tests and I don't think the `Monoid` instance for `Classification` is wrong either...

So either:

1. I don't completely understand the mtl layering and if it has any adverse effects on the total number of tests...or:
2. There is a bug in hedgehog somewhere else

🤷‍♂️

This is the output that hedgehog generates if I add the total to the prints:

```
Test suite test: RUNNING...
━━━ Test.Hedgehog.Text ━━━
  ✓ prop_show_append_size passed 100 tests. Total: 19
  ✓ prop_tripping_append_size passed 100 tests. Total: 10
  ✓ prop_show_append_seed passed 100 tests. Total: 19
  ✓ prop_tripping_append_seed passed 100 tests. Total: 10
  ✓ 4 succeeded.
━━━ Test.Hedgehog.Seed ━━━
  ✓ prop_avoid_pathological_gamma_values passed 1 test. Total: 34
  ✓ 1 succeeded.
━━━ Test.Hedgehog.Classified ━━━
  ✓ prop_check_classifiers passed 1 test. Total: 102
      50.0% big number
      48.0% small number
  ✓ prop_check_coverage passed 1 test. Total: 102
      49.0% big number
      48.0% small number

  ⚠   Only 49.0% big number, but expected 50.0%
      Only 48.0% small number, but expected 50.0%
  ✓ 2 succeeded.
```

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -152,6 +162,55 @@ newtype PropertyT m a =
...
+  (Classification c1 t1) <> (Classification c2 t2) =
+    Classification
+      (Map.foldrWithKey (Map.insertWith (<>)) c1 c2)
+      (t1 + t2)
+
+instance Monoid Classification where
+  mappend =
+    (<>)
+  mempty =
+    Classification mempty 1
```
HuwCampbell: The `1` should be `0`. You start with no tests having been run.

felixmulder: If we set this to zero, we would have to increment the total on each `insertClassifier`. That will only work if the classifiers are total. If we have a set of classifiers that do not cover the complete space, then the total will be wrong. 

In other words, how would we increase the total when no classifier predicate is satisfied?

HuwCampbell: You could also just not record the total here, as the number of tests is available in the report.

This encoding leads to an extra 1 per classification in the test suite AFAICT.

felixmulder: The problem there is that a Property monad is (can be) composed of several smaller Properties - each one run as a leaf in a `TreeT`, so we need to count each node that is hit, not just the whole tree.

See the test output above, it shows it clearer than my sloppy words 😃

HuwCampbell: That's not the purpose of the `TreeT`, or how the `PropertyT` works with it.

The rose tree in `TreeT` is used for shrinking, and yes, while `PropertyT` can be built up monadically, only the *root* (not the leaves) is traversed (or even calculated) in the happy/passing case.

When shrinking get involved, these counters will be hit in a very hard to predict manner, as the shrinking heads up through the binds (PropertyT acts sort of like a non-determinism ListT when shrinking), and would lead to completely skewed results. See #122 on that one.

But it looks like you're not using the `Classification` log from `takeSmallest` so I don't think that's a big deal (you're  `-- FIXME: maybe add classifier to failure?` should probably *not* be done).

felixmulder: Yes you're right, thanks for that!

The problem still stands though, each writer will create a mempty to string along its computation. By changing the definition of mempty I don't know how to inject a count for a non-classified bind.

HuwCampbell: Change Classification to
```haskell
newtype Classification =
  Classification {
      classificationClassifiers :: !(Map ClassifierName Classifier)
    } deriving (Show)
```

`mempty` as an empty map is fine.

And use `reportTests` to determine how good it is a render the percentage.

felixmulder: The total tests in the report does not seem accurate to me - it seems to only count the root Property and not each test inside 🤔

Eg it's 1 for the classified tests above instead of 100 as it should be.

Or am I mistaken?

HuwCampbell: Ahh, right. Your test example is a bit funky.

I would imagine most people are classifying what percentage of their generated values hits certain cases, not generating extra, deterministic values with a `for`.

I might actually go with 1 being the right answer here.

HuwCampbell: Maybe something like this for a deterministic example:

```haskell
prop_check_classifiers :: Property
prop_check_classifiers =
  property $ do
    a <- sized $ pure . unsize
    classify (a < 50) "small number" $
      classify (a >= 50) "big number" success
```

HuwCampbell: Sorry If I'm bring a bit brisk. BTW. I think this is really good.

Regarding the above. Essentially with that `for_` what you're doing is writing that in this single test, "I'm classifying the same state 100 times..." which is different to 100 tests being classified a single time. Or at least I get the feeling it should be.

Does that make sense?

felixmulder: No worries, let's get this right 😀

In that case we should have a way to combine Classifications from different test runs, I'll have a look later this week

felixmulder: No worries, let's get this right 😀

In that case we should have a way to combine Classifications from different test runs, I'll have a look later this week

felixmulder: Once we have that it should be pretty easy to do the beta distribution tho 👍

HuwCampbell: Yep, that's pretty fundamental. On line 199 of runner, it should be `classifier <> classifier0`.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -778,6 +837,49 @@ withRetries :: ShrinkRetries -> Property -> Property
...
+--      for_ xs $ \x ->
+--        classify (x == 0) "newborns" $ do
+--        classify (x > 0 && x < 13) "children" $
+--        classify (x > 12 && x < 20) "teens" $
+--          success
+-- @
+classify :: Bool -> String -> PropertyT IO () -> PropertyT IO ()
+classify = cover 0
+
+cover :: Double -> Bool -> String -> PropertyT IO () -> PropertyT IO ()
```
HuwCampbell: Similar to `annotate` in hedgehog, I think that
```haskell
cover :: (MonadTest m, HasCallStack) => Double -> Bool -> String -> m ()
```
would be a better type signature, and play nicer with how hedgehog tests are usually written.

felixmulder:
Hi all!

I finally had some time to look at this. Been pretty busy.

@HuwCampbell - not sure how I would implement that! If everyone is happy with the changes after this latest round. How about you make that change in a separate PR?

Currently, test output looks something like this:

Screenshot 2019-04-13 at 16 41 50

It's pretty neat!

The tests now only mark a classifier as met, and aggregates the count in the runner. This should give us a better point to do statistics in a follow-up.

Also, this refactor gets rid of the nasty CPP portion that we needed because of GHC 8.2.1!

Cheers,
F

felixmulder:
So I realized that `MonadTest` should just be used as a writer and fixed what @HuwCampbell wanted. So no follow-up on that needed 🎉

felixmulder:
Squashed the history, no more changes unless you have nitpicks 🙏

felixmulder:
@jystic - rebased and ready for your verdict :smiley:

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -110,8 +115,10 @@ import qualified Control.Monad.Trans.Writer.Strict as Strict
 
 import qualified Data.Char as Char
 import           Data.Functor.Identity (Identity(..))
+import           Data.Map (Map)
+import qualified Data.Map as Map
```
jystic: You may want to use the strict map functions here I think.
```haskell
import           Data.Map (Map)
import qualified Data.Map.Strict as Map
```

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -713,6 +776,26 @@ discard :: Monad m => PropertyT m a
...
+--      for_ xs $ \x -> do
+--        classify (x == 0) "newborns"
+--        classify (x > 0 && x < 13) "children"
+--        classify (x > 12 && x < 20) "teens"
+--        success
+-- @
+classify :: (MonadTest m, HasCallStack) => Bool -> String -> m ()
+classify = cover 0
+
+cover :: (MonadTest m, HasCallStack) => Double -> Bool -> String -> m ()
```
jystic: I would like to have a comment for this as it will be visible in the haddock.

@mankyKitty maybe we should mention this in the style guide, public facing functions should have a comment. Internal ones I think it's developer's discretion.

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -33,9 +33,13 @@ import           Control.Monad.Catch (MonadCatch(..), catchAll)
 import           Control.Monad.IO.Class (MonadIO(..))
 
 import           Data.Semigroup ((<>))
+import           Data.Map (Map)
+import qualified Data.Map as Map
```
jystic: As above:
```haskell
import           Data.Map (Map)
import qualified Data.Map.Strict as Map
```

hedgehog/src/Hedgehog/Internal/Property.hs
HuwCampbell: Sorry I think @felixmulder misunderstood my comment.

I suggested importing just Data.Map for the `Map` type.

What @jystic has written is what I was thinking.

felixmulder:
@jystic - done :+1:

hedgehog/src/Hedgehog/Internal/Property.hs
mankyKitty: Sounds good

jystic:
Great, so I started adding a few newtypes around `Int` and `String` and just fixing a few nits before merge and got a bit carried away, spent all day on it but the failure output is much prettier now. @moodmosaic and I spent a long time bikeshedding type names 😆 

An example:
Screenshot 2019-04-21 at 6 06 39 PM

It also prints the current coverage while the test is running, you'll need to run the example to see. I moved the test to `hedgehog-examples` because we actually want them to fail so we can check the output.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -798,6 +939,123 @@ property m =
...
+
+coverageSuccess :: TestCount -> Coverage CoverCount -> Bool
+coverageSuccess tests =
+  null . coverageFailures tests
+
+coverageFailures :: TestCount -> Coverage CoverCount -> [Classifier CoverCount]
+coverageFailures tests (Coverage kvs) =
+  filter (not . classifierCovered tests) (Map.elems kvs)
+
+-- | Records the proportion of tests which satisfy a given condiiton.
```
felixmulder: `condiiton`

jystic: thanks! :D

jystic:
@HuwCampbell I decided I will merge this and do the `Log` integration as a separate PR as it will be easier to review I think. I won't do a release until this happens so I don't think it's a big issue.

jystic:
Semigroup/Monoid 7.10 fun :)

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -820,6 +963,129 @@ property m =
...
+  mappend =
+    (<>)
+
+mkCoverage :: Maybe Span -> ClassifierName -> CoverPercentage -> Cover -> Coverage Cover
+mkCoverage mlocation name minimum_ cover_ =
+  Coverage $
+    Map.singleton name (Classifier name mlocation minimum_ cover_)
+
+coverPercentage :: TestCount -> CoverCount -> CoverPercentage
+coverPercentage (TestCount tests) (CoverCount count) =
```
moodmosaic, line 1022: @felixmulder, couldn't we just write this as

```hs
coverPercentage :: TestCount -> CoverCount -> CoverPercentage
coverPercentage (TestCount tests) (CoverCount count) =
  CoverPercentage (fromIntegral count / fromIntegral tests * 100)
```

What `thousandths` means?

moodmosaic: I'm trying to troubleshoot [this](https://github.com/hedgehogqa/haskell-hedgehog/pull/288#discussion_r940676278), that was the reason I asked.

PR #255, opened 2019-03-20
hedgehog-example: relax resourcet to < 1.3

endgame:
Relax bound, and use `hoist` to work with `MonadUnliftIO`.

See also: discussion about resourcet in #252 .


jystic:
Cool, did you check this actually works?

endgame:
No, I haven't.

endgame:
Okay, I can see the temporary directories appearing and disappearing as the suite runs. I haven't tried to do anything super-crazy like trying to hit it with a signal while it's got a tempdir open.

jystic:
Awesome, thanks!!

HuwCampbell:
I also tested this before suggesting it and dove into the definitions of hoist to assure myself that it would work as expected.

PR #256, opened 2019-03-20
hedgehog:Hedgehog.Internal.HTraversable: Add instance for Proxy

endgame:
For `Command`s that don't need to generate input, it would be possible to use `Proxy`, if it had a `HTraversable` instance, instead of setting up a new data type.


jystic:
Hmm, it will ruin your failure output though, showing `Var 0 <- Proxy` instead of the command name, so it's not really ideal

endgame:
Good point. That's probably enough of a deal-breaker.

PR #257, opened 2019-03-20
Discard on failed pattern matches within a Generator

HuwCampbell:
As discussed in #233.


fosskers:
A similar issue exists for `PropertyT`.

moodmosaic:
See this discussion regarding PropertyT https://github.com/hedgehogqa/haskell-hedgehog/pull/233/files#r234163334

jacobstanley:
I've spent some time thinking about this and the problem I have with it is that ideally one should aim to write tests / generators that don't ever discard.

Discarding slows things down and I would consider its use a bit of an anti-pattern. I think it should only really be used as a last resort. You're better of using the `just` / `filter` functions which locally retry the generation so that your "discard" is limited in scope, rather than throwing away the entire test because some deeply nested generator has failed.

```haskell
just :: MonadGen m => m (Maybe a) -> m a
filter :: MonadGen m => (a -> Bool) -> m a -> m a
```

So I'm just a bit unsure about making it syntactically convenient to discard, as often there are more appropriate options. I actually wonder if it would be better to remove the `MonadFail` instance for `Gen` and have it be a compile error.

moodmosaic:
>I actually wonder if it would be better to remove the MonadFail instance for Gen and have it be a compile error.

This aligns pretty well with @sol's comment on the same topic (QC): https://github.com/nick8325/quickcheck/pull/228#issuecomment-427535945

If we decide to get rid of `MonadFail` perhaps we better do now, before next major release is out.

jacobstanley:
> If we decide to get rid of `MonadFail` perhaps we better do now, before next major release is out.

Too late :sweat_smile:

I think this topic has too many trade-offs to rush the decision, so I'm not that concerned. What we ended up releasing is the same as what it has always been.

HuwCampbell:
I'm going to bump this, as I still think this is a better behaviour.

Issue #250, opened 2019-03-10
Add an example in the haddock for `tripping`


jacobstanley:
```haskell
tripping :: (MonadTest m, Applicative f, Show b, Show (f a), Eq (f a), HasCallStack) => a -> (a -> b) -> (b -> f a) -> m ()
```

The type signature is a bit scary and I don't think it's immediately obvious how to use this.

PR #258, opened 2019-03-20
Add example for tripping

HuwCampbell:
Quick take on #250


jystic:
:+1: thanks, `tripping` has a pretty strange type if you don't know what it's for!

PR #262, opened 2019-03-29
Add labelling of test cases

ruhatch:
This adds `label` and `collect` similar to QuickChecks coverage support. This is achieved using an extra constructor on the existing `Log` type, which collects `Label`s. You could similarly add a `Class` constructor to implement QuickChecks `classify` and I will happily make a PR for that too if this approach seems reasonable.

Display of labels could also use source information to be printed under the `label` command, similar to `annotate`s existing behaviour. If this gets merged I can add that and the other extension mentioned above as issues.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -65,6 +65,10 @@ module Hedgehog.Internal.Property (
   , (===)
   , (/==)
 
+  , label
+  , collect
+  , collectLogLabels
+
   , eval
   , evalM
   , evalIO
@@ -339,6 +343,7 @@ newtype PropertyCount =
 data Log =
     Annotation (Maybe Span) String
   | Footnote String
+  | Label String
     deriving (Eq, Show)
 
 -- | A record containing the details of a test run.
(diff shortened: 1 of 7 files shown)


HuwCampbell:
Thanks for submitting this :)

Just a heads up that it overlaps quite a bit with #253, which is currently not in a fully working state, but is slowly wending through the review process.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -498,6 +503,25 @@ footnoteShow :: (MonadTest m, Show a) => a -> m ()
 footnoteShow =
   writeLog . Footnote . showPretty
 
+-- | Add a label for each test run. It produces a table showing the percentage
+--   of test runs that produced each label.
+--
+label :: MonadTest m => String -> m ()
+label =
```
HuwCampbell: I think this is a better design choice.

moodmosaic: 👍

ruhatch:
@HuwCampbell thanks for the heads up! Looks like #253 wouldn't create too many merge conflicts and covers (haha) the other parts of coverage testing that this PR misses. I think that PR could move the `Classifier`s into a `Log` constructor and then do the `Classifier` monoid operations at the end, which would align it with this approach :man_shrugging:

jystic:
@HuwCampbell I'm going to take over this PR and integrate the two labelling approaches. I've rebased and solved the merge conflicts and will build on top of that.

jystic:
Everything is now being recorded in the log and coverage is calculated prior to reporting.

A screenshot of the coverage example in action:
Screenshot 2019-04-22 at 7 01 53 PM

moodmosaic:
Crazy stuff!

jystic:
I have pondered whether we should hide the coverage results if all conditions are satisfied (the last test `prop_cover_bool`), just to make the output a bit quieter when everything is successful.

moodmosaic:
I think verbose output should be fine in case everything is successful, like in that test for example. It's nice that it shows what the distribution was. Perhaps we can have an option to turn this on/off if needed (RunnerConfig comes to mind).

jystic:
Yeah i was just pondering doing something better than `IO Bool` for test types. It is great for ghci though, which is the main place I use it.

PR #267, opened 2019-04-10
Add 'MonadFail' instance for 'PropertyT'

geigerzaehler:
Add `MonadFail` instance for `PropertyT` so users can write the following tests

    property $ do
      Just val <- lookupUnderTest
      assertSomeProperties val

The above test will produce a test failure when `lookupUnderTest` returns `Nothing`.

We build on the `MonadFail` instances for `TestT` provided by #233.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -624,6 +624,10 @@ instance MonadTrans PropertyT where
   lift =
     PropertyT . lift . lift
 
+instance Monad m => MonadFail (PropertyT m) where
+  fail err =
+    PropertyT (Fail.fail err)
+
 instance MFunctor PropertyT where
   hoist f =
     PropertyT . hoist (hoist f) . unPropertyT


moodmosaic:
FWIW, here's why there isn't a MonadFail instance for `PropertyT`: https://github.com/hedgehogqa/haskell-hedgehog/pull/233#discussion_r234164910 /cc @HuwCampbell

geigerzaehler:
> FWIW, here's why there isn't a MonadFail instance for `PropertyT`: [#233 (comment)](https://github.com/hedgehogqa/haskell-hedgehog/pull/233#discussion_r234164910) /cc @HuwCampbell

Thanks for the hint. I didn’t see that.

As I understand the comments the instance was not added because it was not clear what the behavior should be. I would make the case that having the test failure behavior makes a lot of sense. It is basically `fail = error` (which used to be the default) with more information.

It would be great if the original decision would be reconsidered.

moodmosaic:
>[..]the instance was not added because it was not clear what the behavior should be.

Not quite, I actually think @HuwCampbell makes it clear (in that same thread):

>People expect to see MonadFail errors when upgrading to 8.6. It's part and parcel with the upgrade and that's a good thing. We shouldn't add a `MonadFail` instances just to counter that.

See also https://github.com/nick8325/quickcheck/pull/228#issuecomment-427535945 on a similar discussion in QC space.

jystic:
I certainly do have reservations about fail meaning discard for properties, and even a little for generators. I see properties as advanced tests, rather than generators, so I think error is a sensible choice.

geigerzaehler:
🎉

PR #268, opened 2019-04-11
Make 'PropertyT' an instance of 'MonadResource'

geigerzaehler:
Make `PropertyT` an instance of `MonadResource` simplifies writing tests that use the functions with a `MonadResource` constraint. See the changed example for instance.

### DIFF
--- a/hedgehog-example/src/Test/Example/Resource.hs
+++ b/hedgehog-example/src/Test/Example/Resource.hs
@@ -39,7 +39,7 @@ unixSort input output = do
 
 prop_unix_sort :: Property
 prop_unix_sort =
-  property $ hoist runResourceT $ do
+  property . hoist runResourceT $ do
     values0 <- forAll $
       Gen.list (Range.linear 0 100) $
       Gen.string (Range.constant 1 5) Gen.alpha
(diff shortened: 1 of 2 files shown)


jystic:
This is awesome, I didn't have this from the beggining because, I think I thought ResourceT wouldn't work on the inside, i.e. `hoist` and I wasn't clever enough to get ResourceT to work on the outside of GenT.

See the recent PR https://github.com/hedgehogqa/haskell-hedgehog/pull/255

jystic:
It's funny the only reason the `TestT` / `PropertyT` separation even exists is for `ResourceT` support, maybe I should merge `PropertyT` and `TestT` like it used to be...

geigerzaehler:
Thanks for merging! This is going to make my life a lot easier.

MichaelXavier:
Given that recent versions of `ResourceT` require `MonadUnliftIO`, which forbids instances in stateful monads, is this even possible to use anymore? I've got a test where I'd like to interleave some assertions with a ResourceT so I don't get too far in the test if something is wrong and I'm not seeing a way how. I might have to fall back to removing ResourceT entirely and using `bracket`.

HuwCampbell:
So I believe that the example is now quite dangerous.
```haskell
prop_unix_sort :: Property
property . hoist runResourceT $ do
    (_, dir) <- Temp.createTempDirectory Nothing "prop_dir"
    doStuff
```

If I understand correctly, this will create a directory for every shrink and not clean them up until all shrinking is complete. As shrinking can run hundreds or thousands of times, we will have that many temporary directories on the system; and if it was instead file handles, it could crash the system.

PR #269, opened 2019-04-14
Define proper Applicative instances for Node, Tree and GenT

jystic:
Made the necessary changes to https://github.com/hedgehogqa/haskell-hedgehog/pull/173 so that I was happy to merge.

Not sure about the name `hedgehog-test-laws` as the main point is that it depends on `QuickCheck` but we can always change it later as it's never published anywhere.

The idea would be to switch to using Hedgehog itself when we have some way to verify laws, which requires that we can generate functions.


jystic:
This PR isn't required, I learned how to update the original PR with my changes.

PR #271, opened 2019-04-22
defaultMain: fix build

jystic:


hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -52,12 +51,6 @@ import           Hedgehog.Range (Size)
 
 import           Language.Haskell.TH.Lift (deriveLift)
 
-import           System.Exit (exitFailure)
-#if mingw32_HOST_OS
-import           System.IO (BufferMode (LineBuffering), hSetBuffering, hSetEncoding, stdout, stderr, utf8)
```
moodmosaic: @jystic, turns out this breaks the build on Windows. Just opened https://github.com/hedgehogqa/haskell-hedgehog/pull/273 for it putting back only what was necessary.

PR #272, opened 2019-04-23
 Change Applicative GenT to use zipping

jystic:
This changes the `Applicative` instance for `GenT` so that it [zips](https://github.com/well-typed/integrated-shrinking/blob/master/src/UsingHedgehog.hs#L255-L277) the tree rather than binding it. This means that a generator like `(,) <$> genA <*> genB` will be able to alternate between shrinking `genA` and `genB` rather than doing one first and then the other.

An example shrink tree using a zipping approach vs just using `(<*>) = ap`:

Screenshot 2019-04-23 at 2 35 55 PM

I had it like this originally and changed it because it broke the law `<*> = 
ap`, but I'm not sure this is that important as the `Monad` instance for `GenT` is already unlawful.

[Haxl](https://simonmar.github.io/bib/papers/haxl-icfp14.pdf) also uses an unlawful `Applicative` instance for basically the same reason. This was the motivation for `ApplicativeDo` which you would be able to use with Hedgehog after this patch to achieve more optimal shrinking.

A nice article explaining another option and the tradeoffs: https://elvishjerricco.github.io/2016/09/17/abstracting-async-concurrently.html

I also renamed `Tree` to `TreeT`, so I could have `Tree = TreeT Identity` as I was working with `TreeT Identity` a lot during testing and it got a bit tedious.

### DIFF
--- a/hedgehog-quickcheck/src/Test/QuickCheck/Hedgehog.hs
+++ b/hedgehog-quickcheck/src/Test/QuickCheck/Hedgehog.hs
@@ -8,9 +8,9 @@ import           Control.Monad.Trans.Maybe (runMaybeT)
 import           Data.Functor.Identity (runIdentity)
 
 import           Hedgehog
-import           Hedgehog.Internal.Gen (runGenT)
+import           Hedgehog.Internal.Gen (runGen)
 import qualified Hedgehog.Internal.Seed as Seed
-import           Hedgehog.Internal.Tree (Tree(..), Node(..))
+import           Hedgehog.Internal.Tree (NodeT(..), runTree)
 
 import qualified Test.QuickCheck as QuickCheck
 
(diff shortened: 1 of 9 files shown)


jystic:
hmm can't use CPP `#ifdef` with language extensions, weird

jystic:
It's wild that the scala commits are showing up here, pretty cool

jystic:
I decided requiring `transformers >= 0.5` was a lesser evil than dropping 7.10 support.

7.10 has been a real pain recently though, these last few PRs always seem to have some issue.

It is pretty old now! [Version 7.10.3 (released 8th December 2015)](https://www.haskell.org/ghc/download_ghc_7_10_3.html)

moodmosaic:
I guess we should consider dropping GHC 7.* when GHC 9.* comes out, which might not be so far away given the [increased GHC release cycles](https://www.reddit.com/r/haskell/comments/6qt0iv/ghc_blog_reflections_on_ghcs_release_schedule/).

PR #274, opened 2019-04-24
Cleanup .appveyor.yml

moodmosaic:


moodmosaic:
@jystic I'm waiting to see if build is green, then I'll force push another cleanup (looks like we don't need to install cabal separately)

image

PR #276, opened 2019-04-25
Improved, and dare I say minimal, shrinking of lists

jystic:
This PR provides significantly improved shrinking of lists, I believe Hedgehog should now offer virtually minimal shrinking in this area. The change is that now the list shrink tree can alternate between shrinking the list and shrinking the elements. Before it would shrink all the elements first, and then it would try to shrink the list.

We also simplify `MonadGen` significantly, with all functions being implemented in terms of `toGen` / `fromGen`. This comes at the cost of `StateT` and `RWST` instances, but you can still run these on the inside of the transformer and get a `MonadGen`, it's just slightly less conveniant. If anyone can figure out how to implement `MonadTransControl` for `TreeT` then they will be easy possible on the back of that.

Many thanks to @edsko :100: for the detailed discussions about the improvements offered here, and for his implementation of `interleave` for trees with monadic effects, that is a nut I've been trying to crack for many years now. It is quite a relief to have it solved! 🎉


hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -305,9 +363,20 @@ class Monad m => MonadGen m where
   --
   freezeGen :: m a -> m (a, m a)
 
-instance Monad m => MonadGen (GenT m) where
```
jystic: We remove the `MonadGen` instance for `GenT`, so the combinators in `Hedgehog.Gen` will now only work on `Gen` or transformers including `Gen` as the base monad.

jystic: In the 99% case, users will just use `Gen` by virtue of the type of `forAll` and not worry about any of this.

jystic:
Updated description with details

edsko:
Good stuff :grin:

moodmosaic:
Crazy stuff :rocket:

PR #277, opened 2019-04-27
Use bors on pull requests

justanotherdot:
Hiya :wave: !

Had a chat with @jystic about bringing on bors for in-work Rust port of hedgehog over at  https://github.com/justanotherdot/rust-hedgehog. Seems like integrating it was surprisingly simple and this is the followup PR for integrating it upstream.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -144,5 +144,14 @@ script:
   # Build without installed constraints for packages in global-db
   - rm -f cabal.project.local; ${CABAL} new-build -w ${HC} --disable-tests --disable-benchmarks all
 
+branches:
+  only:
+    # This is where pull requests from "bors r+" are built.
+    - staging
+    # This is where pull requests from "bors try" are built.
+    - trying
+    # Comment this to disable building pull requests.
+    - master
+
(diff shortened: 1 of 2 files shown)


justanotherdot:
Ah, looks like I've left out appveyor, I'll add that in, too.

moodmosaic:
Cool! Should we get the Rust port under Hedgehog QA at some point

justanotherdot:
> Cool! Should we get the Rust port under Hedgehog QA at some point

yeah, still chipping away at it slowly :sweat_smile: 

Looks like we may need to troubleshoot a tiny bit here. The comment Jacob made should have triggered a run for bors plus merged the PR on success and deleted the branch, so not sure what's up. I'll let Jacob. See justanotherdot/rust-hedgehog#20 for an example

jystic:
ah, was stuck on the enter password page

justanotherdot:
I wonder if we need to kick it with an empty commit or something if it’s
now integrated?

jystic:
maybe I need to merge so it's on master?

justanotherdot:
nice, he's alive!

PR #278, opened 2019-04-27
Drop support for GHC 7.10

jystic:
I keep running up against compatibility issues, and I think it's old enough now.

We'll still have support for GHC 8.01, which is three years old now.


jystic:
bors try

PR #280, opened 2019-04-28
Revert .travis.yml too

justanotherdot:


jystic:
Will integrate this as part of #278

Issue #281, opened 2019-04-28
Improve `filter`


edsko:
At the moment, `filter` is defined like

```haskell
filter :: MonadGen m => (a -> Bool) -> m a -> m a
filter p gen =
  let
    try k =
      if k > 100 then
        discard
      else do
        x <- scale (2 * k +) gen
        if p x then
          pure x
        else
          try (k + 1)
  in
    try 0
```

I think this is not quite right. The problem is that this is using the monadic interface to the generator; therefore the check `if p x then ... else ..` gets applied _at every level of the shrink tree for `x`_. That's not correct: that means that the entire generator is re-run every time a shrunk value happens not to satisfy the predicate, at which point we'd basically start over. This leads to large shrink trees and consequently bad performance (https://stackoverflow.com/questions/54412108/why-the-does-this-shrink-tree-looks-the-way-it-does-when-using-filter). Indeed, it means that we may end up "shrinking" a value to a larger one.

Instead what should happen after we find a value that satisfies the predicate, we should _remove_ any shrunk values that don't satisfy the predicate. Just as a proof of concept, this is easily defined in terms of the existing filter:

```haskell
filter' :: MonadGen m => (a -> Bool) -> m a -> m a
filter' p gen = Gen.Internal.ensure p $
                  join (snd <$> Gen.filter (p . fst) (Gen.freeze gen))
```

(Not entirely sure why `ensure` is not exported?). I'm calling this a proof of concept because what should probably happen instead is that `filter` should be _changed_ to behave like `filter'` (or at least the existing `filter` should not be the default).

The definition of `filter'` is not quite ideal. The problem is that `ensure` will _stop_ as soon as it finds one shrunk value that doesn't satisfy the predicate, rather than skipping to the next level in the tree. However, QuickCheck does this too; given

```haskell
newtype Uniform = Uniform Int
  deriving (Show)

instance Arbitrary Uniform where
  arbitrary = Uniform <$> choose (0, 100)
  shrink (Uniform n) = Uniform <$> shrink n
```

we get

```
> quickCheck (\(Uniform n) -> even n)
*** Failed! Falsified (after 1 test):                  
Uniform 31
```

Nonetheless, I still don't think that it makes much sense to re-run the entire generator when shrinking happens to hit on an element that doesn't satisfy the predicate.

# Variant for better shrinking

(The below is independent from the bugfix above.)

[...]

PR #282, opened 2019-04-29
Fix filter shrinking bug #281

jystic:
This implements @edsko's ideas from #281. I made the default `filter` the "ideal" one presented last in that issue, it just requires that the base `Monad` is `Identity` which I would expect to be the case 100% of the time when using `filter`. The faster one is `filterT` and can be used with any base `Monad`, I'm not sure if this should even be included. It is indeed faster but i'm not sure if that's so important?

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -95,7 +95,9 @@ module Hedgehog.Internal.Gen (
   , discard
   , ensure
   , filter
+  , filterT
   , just
+  , justT
 
   -- ** Collections
   , maybe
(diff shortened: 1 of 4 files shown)


jystic:
Ah so actually, having `StateT` on the inside of `GenT` means you need to use `filterT` because the `GenBase` is `StateT s Identity` and not `Identity`, so we might need to keep it. It is still pure though so I wonder if there's something we can do to make that possible.

Having `StateT` on the outside of `GenT` means it isn't a `MonadGen` so we're stuck there too.

jystic:
Going to merge this, if you have any comments I'm happy to integrate them in a follow up, I have some more PRs based on this.

edsko:
Nice work! :) I think it's useful to keep `filterT` around; the ideal `filter` is great most of the time, but when it's expensive to look for elements that do not satisfy a predicate, we may indeed wish to stop as soon as we find such an element and not continue, in order to keep the cost of shrinking down. It's also hard to define the equivalent of `filter` in a type class based approach to shrinking (like QuickCheck) and so I guess we simply have less experience with it as a community.

PR #283, opened 2019-04-29
Fixed state machine duplicate Vars when shrinking

jystic:
Running a `StateT` on the inside of `GenT` doesn't work properly with the new applicative implementation, so we `distributeT` to move it to the outside before running it.

It has to be on the inside, because otherwise it's not a `MonadGen`, due to the new list shrinking code. It does seem kind of odd that I can't make it work on the outside, it must be possible.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/State.hs
+++ b/hedgehog/src/Hedgehog/Internal/State.hs
@@ -71,6 +71,7 @@ import qualified Data.Map as Map
 import qualified Data.Maybe as Maybe
 import           Data.Typeable (Typeable, TypeRep, Proxy(..), typeRep)
 
+import           Hedgehog.Internal.Distributive (distributeT)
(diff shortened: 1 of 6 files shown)


hedgehog/src/Hedgehog/Internal/State.hs
```diff
@@ -572,7 +573,7 @@ genActions ::
   -> Context state
   -> gen ([Action m state], Context state)
 genActions range commands ctx = do
-  xs <- Gen.fromGenT . hoist (`evalStateT` ctx) $ Gen.list range (action commands)
+  xs <- Gen.fromGenT . (`evalStateT` ctx) . distributeT $ Gen.list range (action commands)
```
jystic: in this context:
```haskell
distributeT :: GenT (StateT (Context state) Identity) a -> StateT (Context state) (GenT Identity) a
```

PR #288, opened 2019-05-11
Statistically verify result

felixmulder:
This PR adds ability to be able to statistically verify that a result is correct. It does not _yet_ add the ability to cancel running tests if they can never reach the specified coverage.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -1068,12 +1068,10 @@ confidenceSuccess tests confidence =
     -- making the API less intuitive
     tolerance = 0.9
     assertLow :: Label CoverCount -> Bool
-    assertLow MkLabel{..} =
-      wilsonLowerBound
-        (fromIntegral $ unCoverCount labelAnnotation)
-        (fromIntegral tests)
-        (1 / fromIntegral (unConfidence confidence))
-      >= tolerance * (unCoverPercentage labelMinimum / 100.0)
+    assertLow coverCount@MkLabel{..} =
+      fst (boundsForLabel tests confidence coverCount)
+      >=
+      tolerance * (unCoverPercentage labelMinimum / 100.0)
   in
     and . fmap assertLow . Map.elems . coverageLabels
 
@@ -1083,36 +1081,37 @@ confidenceFailure :: TestCount -> Confidence -> Coverage CoverCount -> Bool
 confidenceFailure tests confidence =
   let
     assertHigh :: Label CoverCount -> Bool
-    assertHigh MkLabel{..} =
-      wilsonUpperBound
-        (fromIntegral $ unCoverCount labelAnnotation)
-        (fromIntegral tests)
-        (1 / fromIntegral (unConfidence confidence))
-      < (unCoverPercentage labelMinimum / 100.0)
+    assertHigh coverCount@MkLabel{..} =
+      snd (boundsForLabel tests confidence coverCount)
+      <
+      (unCoverPercentage labelMinimum / 100.0)
   in
     or . fmap assertHigh . Map.elems . coverageLabels
+
+boundsForLabel :: TestCount -> Confidence -> Label CoverCount -> (Double, Double)
+boundsForLabel tests confidence MkLabel{..} =
+  wilsonBounds
+    (fromIntegral $ unCoverCount labelAnnotation)
+    (fromIntegral tests)
+    (1 / fromIntegral (unConfidence confidence))
 
 -- In order to get an accurate measurement with small sample sizes, we're
 -- using the Wilson score interval
 -- (<https://en.wikipedia.org/wiki/Binomial_proportion_confidence_interval#Wilson_score_interval
 -- wikipedia>) instead of a normal approximation interval.
-wilson :: Integer -> Integer -> Double -> Double
-wilson k n z =
+wilsonBounds :: Integer -> Integer -> Double -> (Double, Double)
+wilsonBounds k n a =
   let
-    nf = fromIntegral n
     p = fromIntegral k / fromIntegral n
+    nf = fromIntegral n
+    wilson z =
+      (p + z * z/(2 * nf) + z * sqrt (p * (1 - p) / nf + z * z / (4 * nf * nf)))
+      /
+      (1 + z * z / nf)
+    low = wilson . invnormcdf $ 1 - (1 - a) / 2
+    high = wilson . invnormcdf $ 1 - a / 2
   in
-    (p + z * z/(2 * nf) + z * sqrt (p * (1 - p) / nf + z * z / (4 * nf * nf)))
-    /
-    (1 + z * z / nf)
-
-wilsonLowerBound :: Integer -> Integer -> Double -> Double
-wilsonLowerBound k n a =
-  wilson k n $ invnormcdf (1 - (1 - a / 2))
-
-wilsonUpperBound :: Integer -> Integer -> Double -> Double
-wilsonUpperBound k n a =
-  wilson k n $ invnormcdf (1 - a / 2)
+    (low, high)
 
 fromLabel :: Label a -> Coverage a
 fromLabel x =
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -31,6 +31,7 @@ module Hedgehog.Internal.Property (
   , ShrinkLimit(..)
   , ShrinkCount(..)
   , ShrinkRetries(..)
+  , withConfidence
   , withTests
   , withDiscards
   , withShrinks
@@ -221,6 +222,15 @@ newtype PropertyName =
       unPropertyName :: String
     } deriving (Eq, Ord, Show, IsString, Semigroup, Lift)
 
+-- | The acceptable occurrence of false positives
+--
+--   Example, @Confidence 10^9@ would mean that you'd accept a false positive
+--   for 1 in 10^9 tests.
+newtype Confidence =
+  Confidence {
+    unConfidence :: Int
+  } deriving (Eq, Ord, Show, Num, Lift)
+
 -- | Configuration for a property test.
 --
 data PropertyConfig =
@@ -229,6 +239,7 @@ data PropertyConfig =
     , propertyDiscardLimit :: !DiscardLimit
     , propertyShrinkLimit :: !ShrinkLimit
     , propertyShrinkRetries :: !ShrinkRetries
+    , propertyConfidence :: !(Maybe Confidence)
     } deriving (Eq, Ord, Show, Lift)
 
 -- | The number of successful tests that need to be run before a property test
@@ -898,6 +909,8 @@ defaultConfig =
         1000
     , propertyShrinkRetries =
         0
+    , propertyConfidence =
+        Nothing
     }
 
 -- | Map a config modification function over a property.
@@ -905,6 +918,10 @@ defaultConfig =
 mapConfig :: (PropertyConfig -> PropertyConfig) -> Property -> Property
 mapConfig f (Property cfg t) =
   Property (f cfg) t
+
+withConfidence :: Confidence -> Property -> Property
+withConfidence c =
+  mapConfig $ \config -> config { propertyConfidence = Just c }
 
 -- | Set the number of times a property should be executed before it is considered
 --   successful.
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -103,6 +103,7 @@ module Hedgehog.Internal.Property (
   -- $internal
   , defaultConfig
   , mapConfig
+  , allowsEarlyStop
   , failDiff
   , failException
   , failWith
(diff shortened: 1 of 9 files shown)


felixmulder:
Eum, turned out to be trivial to add support for aborting a test that cannot satisfy coverage. So added that :tada:

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -174,14 +191,28 @@ checkReport cfg size0 seed0 test0 updateUI =
...
+            seed
+            0
+            (Just coverage0)
+            Nothing
+            "Test coverage cannot be reached, aborted"
+            Nothing
+            []
+
       else if tests >= fromIntegral (propertyTestLimit cfg) then
         -- we've hit the test limit
```
felixmulder: One thing to note here is that just as we can stop early if we'll never reach the desired coverage - we *can* also stop early if we've already reached it!

I guess that is sort of a design decision - should we stop if we've already reached the desired coverage?

felixmulder: Actually - turned out that there was a bug hiding if we didn't do this. Fixed it - I'll give this a rest now.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -221,6 +229,15 @@ newtype PropertyName =
       unPropertyName :: String
     } deriving (Eq, Ord, Show, IsString, Semigroup, Lift)
 
+-- | The acceptable occurrence of false positives
+--
+--   Example, @Confidence 10^9@ would mean that you'd accept a false positive
+--   for 1 in 10^9 tests.
+newtype Confidence =
+  Confidence {
+    unConfidence :: Integer
```
jacobstanley: Do you envision using this with confidence greater than `10^18`? (`Int64`)

jacobstanley:
:chart_with_upwards_trend:  All looks pretty legit, although the math/stats is a bit beyond me

@HuwCampbell if you're around, would you mind having a look at this?

hedgehog/src/Hedgehog/Internal/Property.hs
felixmulder: QuickCheck gives a pretty good explanation of this in their docs:

> If you are using checkCoverage as part of a test suite, you should be careful not to set certainty too low. If you want, say, a 1% chance of a false positive during a project's lifetime, then certainty should be set to at least 100 * m * n, where m is the number of uses of cover in the test suite, and n is the number of times you expect the test suite to be run during the project's lifetime. The default value is chosen to be big enough for most projects.

So for instance - let's say we're writing a big project with 100 ocurrences of `cover`, then we'd get:

```
100 * 100 * n = confidence
```

Let's say that we expect to run the project 20 times per day per developer with 5 developers. That would make it:

```
100 * 100 * (20 * 5 * daysWithoutFalsePositives) = confidence
<=>
1 000 000 * daysWithoutFalsePositives = confidence
```

Let's say we want 5 years with 1% chance of false positives:

```
1 000 000 * daysWithoutFalsePositives = confidence
<=>
1 000 000 * 365 * 5 = 1 825 000 000
```

Which is veeeeeeery roughly `2*10^9`. So yeah, maybe not. Would you rather use `Int64`?

hcampbell-tyro:
> @HuwCampbell if you're around, would you mind having a look at this?

I'll try and have a look before the weekend.

HuwCampbell:
Sorry I don't think the statistical reasoning is correct, you appear to have mixed different metrics inappropriately.

I personally would not have gone with a Wilson metric (but whatever really), and used a Beta distribution prior to approximate the confidence interval.

From a user experience note, I think the conditionals would be better off a different way as well.

Your example has this
```haskell
prop_without_confidence :: Property
prop_without_confidence =
  withConfidence (10^9) . withTests 1000000 . property $ do
    number <- forAll (Gen.int $ Range.linear 1 10)
    cover 60 "number == 1" $ number == 1
```

To me this says, run the test at the very least 1000000 times (because I want to check it that well) while also ensuring confidence of 10^9.

But the way you've done the short circuit breaks that minimum test requirement, it could essentially be over in 50 tests, which isn't cool IMO.

felixmulder:
When it comes to the amount of tests run - we can simply change the implementation to adhere to `withTests X` to run and present the certainty after it's run. We might also want to have a mode to run "until confidence fulfilled". WDYT?

The current implementation simply mimics what QC does for its confidence tests - totally open to changing that.

When it comes to the Wilson score - from what I can tell, it's an improvement to the normal approximation especially when run with lower amount of sampling. But I'm no expert. Since it's whatever for you - then I propose we keep as is.

In the end, I'm not sure what your review boils down to when saying "I don't think the statistical reasoning is correct".

What changes would you like to see in order for us to make progress on this? :)

felixmulder:
```haskell
-- run tests until confidence has been reached, or fail if cannot be reached:
prop_p1 =
  withConfidence (10^9) . property $ do
    ...

-- run at least X tests and present confidence:
prop_p2 =
  withTests X . property $ do
    ...

-- run at least X times and present confidence, failing if test hasn't reached it:
prop_p3 =
  withConfidence (10^9) . withTests X . property $ do
    ...
```

Example behaviors :point_up:

HuwCampbell:
You're absolutely right that wasn't very helpful, my apologies. I'll try to come back to this tonight and give more concrete thoughts.

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -1021,6 +1048,58 @@ coverageFailures :: TestCount -> Coverage CoverCount -> [Label CoverCount]
...
+    nf = fromIntegral n
+    p = fromIntegral k / fromIntegral n
+  in
+    (p + z * z/(2 * nf) + z * sqrt (p * (1 - p) / nf + z * z / (4 * nf * nf)))
+    /
+    (1 + z * z / nf)
+
+wilsonLowerBound :: Integer -> Integer -> Double -> Double
+wilsonLowerBound k n a =
+  wilson k n $ invnormcdf (a / 2)
```
HuwCampbell: I don't think this is the right calling convention for `wilson`. `z` is the probit for the error bounds, and is included in the denominator of wilson as `z^2`. I think you still need to use `1  - a / 2`, but with a negative sign, in `wilson`, not a plus.

This is throwing off the lower bound by quite a margin, especially with confidence intervals around 10^9.

Comparing with a beta distribution with alpha 5, beta 10, and 10^7, it's 0.045 when it should be more like 0.008.

hedgehog/src/Hedgehog/Internal/Property.hs
HuwCampbell: Yeah we're talking age of the universe scales at 10^18.

HuwCampbell:
> It's quite plausible that we'd see the random walk spuriously affecting the results, as we're not taking multiple measurements into account (the metrics include the implicit assumption that we're only testing at the end of the trial).

It looks like quickcheck somewhat sidesteps this by only running the statistical test on tests 200, 400, 800, etc.

I had a look at QC's code... I'll think about it harder but I do believe they've misimplemented their lower bound calculation as well.

HuwCampbell:
Ok, I believe I have figured out what's going on.

I still think that the way that the Wilson function is being called is incorrect (or at least extremely counter-intuitive), but the symmetry of `invnormcdf` around 0.5 cancels it out completely.

I would personally like to see the type signature changed to
```
wilsonBounds :: Integer -> Integer -> Double -> (Double, Double)
```
with the plus/minus made explicit.

After that if we tackle the early stopping problem I think this will be in good shape.

felixmulder:
Cool! Thanks for the comments @HuwCampbell - I've been pretty busy, but I guess we're in no rush. I'll try to get to this some time next week :)

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -1021,6 +1048,58 @@ coverageFailures :: TestCount -> Coverage CoverCount -> [Label CoverCount]
...
+confidenceSuccess :: TestCount -> Confidence -> Coverage CoverCount -> Bool
+confidenceSuccess tests confidence =
+  let
+    assertLow :: Label CoverCount -> Bool
+    assertLow MkLabel{..} =
+      wilsonLowerBound
+        (fromIntegral $ unCoverCount labelAnnotation)
+        (fromIntegral tests)
+        (1 / fromIntegral (unConfidence confidence))
+      >= 0.9 * (unCoverPercentage labelMinimum / 100.0)
```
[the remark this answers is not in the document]
felixmulder: This is just a tolerance - we could include that in the confidence data type. That would be clearer.

felixmulder:
@HuwCampbell: took care of the early stops, now it works as advertised above. I.e:

- If *only* `withTests n` is set, exactly `n` tests will be run
- If *only* `withConfidence c` is set, tests will be run until the confidence is met or known to never be met
- If *both* `withTests n` *and* `withConfidence c` are used, *at least* `n` tests will be run, but the suite will continue until the confidence is reached or deemed to be unreachable


With regards to changing the signature to:

```haskell
wilsonBounds :: Integer -> Integer -> Double -> (Double, Double)
```

I'm trying to grok a how you'd like that to look, would the returned tuple be the lower and upper bound? If so - the implementation doesn't really use them together (therefore the split as `wilsonLowerBound` and `wilsonUpperBound`).

However, I made the negative sign explicit :)

HuwCampbell:
> I'm trying to grok a how you'd like that to look, would the returned tuple be the lower and upper bound? If so - the implementation doesn't really use them together (therefore the split as `wilsonLowerBound` and `wilsonUpperBound`).

Yes, that's what I was thinking, I don't think it matters that they're not used together, Haskell is lazy, so it won't do extra work.

felixmulder:
> Haskell is lazy, so it won't do extra work.

Fair point 😄 

Should be gtg now then!

felixmulder:
Ping pong @HuwCampbell :)

patrickt:
Would love to see this land! ❤️

felixmulder:
Alright - CI seems happy now.

The functionality now behaves like this:

- `withConfidence` adds confidence checking to the result of test run (but obey the `withTests minTests` parameter
- `verifiedTermination` will run at least 100 tests and then check every 100 tests if success is verified or if failure is verified, this means that is might run less than the tests specified with `withTests`, but at least 100 tests for sure

felixmulder:
Oh my :sob: :smile:  :tada:

patrickt:
Yay!

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -221,14 +237,23 @@ newtype PropertyName =
...
+newtype Confidence =
+  Confidence {
+    unConfidence :: Int64
+  } deriving (Eq, Ord, Show, Num, Lift)
+
 -- | Configuration for a property test.
 --
 data PropertyConfig =
   PropertyConfig {
-      propertyTestLimit :: !TestLimit
```
symbiont-sam-halliday: removing this field broke the API between 1.0.1 and 1.0.2 :disappointed:

felixmulder: Where's MIMA when you need it 😢

moodmosaic: Was it removed by accident? Also—it's time ti start adding some more automated tests.

HuwCampbell: I think everything in `Hedgehog.Internal.Property` is considered internal.

moodmosaic: Was it removed for a reason though? (We *do* try not to break other peoples code if possible.)

hedgehog/src/Hedgehog/Internal/Property.hs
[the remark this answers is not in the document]
moodmosaic: @felixmulder, @HuwCampbell, perhaps there's something off with the above calculation? See test below:

```hs
prop_with_confidence :: Property
prop_with_confidence =
  verifiedTermination . withConfidence (10^9) . withTests 1000000 . property $ do
    number <- forAll (Gen.int $ Range.constant 1 2)
    cover 50 "number == 1" $ number == 1
    cover 50 "number == 2" $ number == 2
``

This will run forever, even though the actual coverage has been reached.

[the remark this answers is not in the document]
moodmosaic: I understand that this snippet is outdate, but to me it looks that it's around this part that the issue happens.

`confidenceSuccess` and `confidenceFailure`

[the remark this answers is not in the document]
moodmosaic: @felixmulder @HuwCampbell 🏓 🍻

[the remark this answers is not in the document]
HuwCampbell: The "whole" confidence interval must be greater than the desired coverage. Here, the more tests we run just shrinks the interval from like 49.9% to 50.1% to 49.99% to 50.01%.

With more tests we just get more 9s.

Put another way, the cover query states that the _lower_ bound must be greater than the desired coverage, which it's clearly not. The 49.99999% is never getting above 50%.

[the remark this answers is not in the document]
moodmosaic: It almost looks as QuickCheck's [`checkCoverage`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:checkCoverage) (when using [`cover`](https://hackage.haskell.org/package/QuickCheck-2.14.2/docs/Test-QuickCheck.html#v:cover)) is far less stringent in order to improve performance.

Here's a [side-by-side comparison](https://github.com/moodmosaic/coverage-check-example/tree/master/src/Test/Q72168364) of Hedgehog and QuickCheck with regards to this. If I use the exact same percentage, Hedgehog runs forever, while QuickCheck terminates.

@HuwCampbell, I would be very much interested in your thoughts on this. Thank you.

PR #289, opened 2019-05-11
Add histograms to labels / coverage

jystic:
This PR adds unicode histograms / percentage bars to the new labelling / coverage system. Not sure if I have quite the right balance yet, but time will tell.

For labels with coverage requirements, the minimum required coverage is shown on the right hand side of the histogram, along with a tick or a cross indicating whether we met the requirement. We also highlight visually (with red dots) the additional coverage required to make it to the requirement, not sure if I like this yet but we'll give it go.

Shame :australia: :greece: haven't won any world cups :cry:

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -8,6 +8,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections #-}
+
 module Hedgehog.Internal.Report (
   -- * Report
     Summary(..)
@@ -209,6 +210,7 @@ data Markup =
   | SuccessText
   | CoverageIcon
   | CoverageText
+  | CoverageFill
   | DeclarationLocation
   | StyledLineNo !Style
   | StyledBorder !Style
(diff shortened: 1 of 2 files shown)


hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -869,6 +906,93 @@ ppCoverPercentage :: CoverPercentage -> Doc Markup
...
+      --     ++ " " ++ show fillErrorWidth
+      --     ++ " " ++ "+"
+      --     ++ " " ++ show fillSurplusWidth
+      --     ++ " " ++ "="
+      --     ++ " " ++ show fillWidth
+      ]
+  in
+    bar ('█', ['·', '▏', '▎', '▍', '▌', '▋', '▊', '▉'])
+
+    -- FIXME Maybe this should be configurable?
```
moodmosaic, line 987: Indeed, it might be a good idea to have this configurable, @jacobstanley.

I've noticed that on (older) Windows machines without WSL/WSL2 and/or earlier versions of Windows 10 2004 (build 19041) the current bars won't render correctly (the last one, basically, `'▉'`).

OTOH, this renders just fine:

```hs
--bar ('⣿', ['·', '⡀', '⡄', '⡆', '⡇', '⣇', '⣧', '⣷'])
```

PR #300, opened 2019-05-30
Add basic symbol rendering for non-Unicode terminals

ag-eitilt:
The [Windows Unicode workaround](https://github.com/hedgehogqa/haskell-hedgehog/blob/bfd9b93e328104e6273b5b1bd427c2789b769cfd/hedgehog/src/Hedgehog/Internal/Report.hs#L1171-L1175) is certainly valid for that platform, but it doesn't apply everywhere; case in point, Hedgehog crashes on my standard terminal (raw Linux tty) with the infamous `commitBuffer: invalid argument (invalid character)` message, since I keep it on a Latin-1 encoding for font support.  This fixes that for any ASCII-compatible encoding.

I'm not entirely happy with some of the symbols (there's *no* way to get a good checkmark in ASCII), and am completely happy to change any of my reassignments, but the test suite output looks about as good as I'm expecting it to.  There's still some notable text that doesn't get shown there (eg. [`Tree`](hedgehog/src/Hedgehog/Internal/Tree.hs) and [`Tripping`](hedgehog/src/Hedgehog/Internal/Tripping.hs)) so I'm not completely sure those fit in with everything else, though; they should -- they use the same design language -- but I can't guarantee it.

EDIT: One last thing I should mention: this doesn't update if the locale is changed after the test runner is started.  I don't see that as a large failing in as short-lived an application as they tend to be, and not worrying about it allows `unicode` to be used outside `MonadIO` which *is* a notable help.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -638,21 +638,21 @@ failDiff x y =
         failWith Nothing $
         unlines $ [
             "Failed"
-          , "━━ lhs ━━"
+          , unicode "━━ lhs ━━" "-- lhs --"
           , showPretty x
-          , "━━ rhs ━━"
+          , unicode "━━ rhs ━━" "-- lhs --"
           , showPretty y
           ]
 
     Just vdiff@(ValueSame _) ->
       withFrozenCallStack $
         failWith (Just $
-          Diff "━━━ Failed ("  "" "no differences" "" ") ━━━" vdiff) ""
+          Diff (unicode "━━━ Failed (" "=== Failed (")  "" "no differences" "" (unicode ") ━━━" ") ===") vdiff) ""
 
     Just vdiff ->
       withFrozenCallStack $
         failWith (Just $
-          Diff "━━━ Failed (" "- lhs" ") (" "+ rhs" ") ━━━" vdiff) ""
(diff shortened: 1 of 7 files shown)


jacobstanley:
Have been pondering this for a few days, just though I should check in :)

I think having this feature is a good idea, but I'm not thrilled about how invasive it is, have been trying to think of alternative designs, but all have pros and cons. One option I was thinking was a lookup table for say "replace `━` with `-`" which we could just apply to the entire output.

Still giving it some thought :sweat_smile:

ag-eitilt:
Oh, definitely!  I wasn't quite happy with that method either, but had 
the same trouble coming up with alternates.  Eventually I went with 
"anything is better than nothing" in making it seem like I wasn't just 
complaining, but if you can figure out something better, go ahead with 
my blessing.

One benefit of a lookup table is that several symbols wind up being 
reused several times over, and a central registry like that would help 
with DRY.

Another thing that might work, but which would probably wind up 
initially being even more invasive, would be integrating some l10n/i18n 
library.  With the issue being the output not matching the locale, if we 
can leverage the same mechanism for detecting if the console supports 
Unicode as we do for detecting whether we need to print in Spanish, the 
complexity might wind up being mitigated a bit.  I'm just not sure if 
there's any good libraries for that out there; I've not looked too 
carefully, but I've not been impressed by what I've seen so far.

PR #305, opened 2019-06-17
Added a SizeLimit configuration option

etorreborre:
and bumped other "limit" options to Int64.

Hey there. I am currently working on integrating "enumerative testing" to Hedgehog where, instead of randomly selecting values out of a given range we enumerate them exhaustively for small size (ala SmallCheck) then switch to random selection when there are too many combinations. I think that this is a way to get a nice coverage, especially of "business" data structures which are not recursive. 

However in order to integrate those `Enumerates` (coming from the `testing-feat` package) as generators we need to be able to "index" into them. However the only two parameters passed to a `GenT` are `Size` and `Seed`. I would like to use `Size` for my indexing into an enumeration so I need it to be able to go from `0` to a configured value.

So I am proposing to be able to have the currently hard-coded `99` size value as a configuration option, which shouldn't hurt any current user. At the same time, those enumerations can be quite large so the size needs to be bounded and `Int64` is a good candidate for that. I know this has some memory implications but I don't think that they matter much for test execution.

Here is an example of how I am using all of this to enumerate the values of a datatype: https://github.com/etorreborre/registry-hedgehog/blob/testing-feat/test/Test/Data/Registry/TestingFeatSpec.hs#L39

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -1,4 +1,4 @@
-version: 1.0
+version: 1.0.1
 
 name:
   hedgehog
--- a/hedgehog/src/Hedgehog.hs
+++ b/hedgehog/src/Hedgehog.hs
@@ -75,6 +75,9 @@ module Hedgehog (
   , withShrinks
   , ShrinkLimit
 
+  , withSize
+  , SizeLimit
+
   , withRetries
   , ShrinkRetries
 
@@ -168,6 +171,7 @@ import           Hedgehog.Internal.Property (LabelName, MonadTest(..))
 import           Hedgehog.Internal.Property (Property, PropertyT, PropertyName)
 import           Hedgehog.Internal.Property (Group(..), GroupName)
 import           Hedgehog.Internal.Property (ShrinkLimit, withShrinks)
+import           Hedgehog.Internal.Property (SizeLimit, withSize)
 import           Hedgehog.Internal.Property (ShrinkRetries, withRetries)
 import           Hedgehog.Internal.Property (Test, TestT, property, test)
 import           Hedgehog.Internal.Property (TestLimit, withTests)
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -139,7 +139,6 @@ import qualified Control.Monad.Trans.State.Strict as Strict
 import qualified Control.Monad.Trans.Writer.Lazy as Lazy
 import qualified Control.Monad.Trans.Writer.Strict as Strict
 
-import           Data.Int (Int64)
 import qualified Data.Char as Char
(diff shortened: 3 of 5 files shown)


jacobstanley:
> At the same time, those enumerations can be quite large so the size needs to be bounded and `Int64` is a good candidate for that.

fwiw `Int` = `Int64` on basically every platform, I think it's even 64-bit on windows now

`Int` is also bounded

symbiont-eric-torreborre:
Of course, I'm silly. I will fix that in the PR.

Otherwise what do you think of this Size parameter?

symbiont-eric-torreborre:
Removed the `Int64` in a fixup commit.

etorreborre:
> Otherwise what do you think of this Size parameter?

@jacobstanley ping, in case you didn't see my question

jacobstanley:
I'm nervous about possible side effects that we're not aware of, but I suppose it's opt-in as you point out

All the generators assume size is only from 0-99 so i'm slightly concerned it may have unintended side effects? Not sure how we can check everything doesn't go crazy! But yeah hopefully we can get this through.

So, if you had access to just the current test number would that also solve your issue? I actually wanted this the other day.

symbiont-eric-torreborre:
> So, if you had access to just the current test number would that also solve your issue? 

If that was accessible as a parameter to `runGenT` in addition to size and seed yes that would work.

symbiont-eric-torreborre:
This might be a bigger change in the code base though? Also I'm curious, what was your use case?

symbiont-eric-torreborre:
@jacobstanley If you're onboard with this idea, I can close this PR and start another one.

jacobstanley:
Yeah that is a huge change! I didn't think of that. Hmm, not sure I'm ready to commit yet but it would be handy.

I want to add golden testing to hedgehog and I want to produce a different output file for each test. I could use the size, but that would limit me to 100 files, which is probably plenty.

symbiont-eric-torreborre:
Actually `runGenT` is used 19 times in `Gen.hs` and `GenT` 13 times. It seems that it's all located to one file.  This might not be a big deal, except maybe the lack of backward compatibility (at the same time this is an internal API).

etorreborre:
https://github.com/etorreborre/haskell-hedgehog/tree/gen-with-tests-number

@jacobstanley if you want to get a feel for the changes involved in using the number of tests in a generator. Note that:

 - I had to put the `TestCount` into its own file
 - I arbitrarily passed `0` to `evalGen`, maybe we should add another function `evalGenAt :: TestCount ...` taking the tests number as a parameter as well

BTW github won't let me create a PR for now, error 500 👎

etorreborre:
The PR for using the test count is here: https://github.com/hedgehogqa/haskell-hedgehog/pull/307

etorreborre:
I'm going to close this PR for now and just leave the other PR where we can access the test count.

PR #307, opened 2019-06-20
use the current number of tests in a generator

etorreborre:
@jacobstanley if you want to get a feel for the changes involved in using the number of tests in a generator. Note that:

 - I had to put the `TestCount` into its own file
 - I arbitrarily passed `0` to `evalGen`, maybe we should add another function `evalGenAt :: TestCount ...` taking the tests number as a parameter as well

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,8 +25,8 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: haskell/actions/setup@v1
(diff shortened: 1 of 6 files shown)


etorreborre:
Mazel tov, I was able to create the PR!

etorreborre:
Hi @jacobstanley did you have time to give a bit more thoughts to this PR?

etorreborre:
Hi @jacobstanley, better late than never I updated that PR if you still think this could be a good idea.

etorreborre:
Ok, I'm going to close this PR now so that it disappears from my backlog. Please re-open if you feel like it.

PR #308, opened 2019-06-20
Include location of failed assertion in report

### LINKED ISSUE #306: Include file/line/column of failed assertion in output?
Hi! This is going to sound like a very Emacs-specific problem, but please bear with me. :slightly_smiling_face: 

I'm using Hedgehog a lot through Emacs' _compilation-mode_ (or the _haskell-compilation_ buffer, to be specific). I was trying to add a pattern to `haskell-compilation-error-regexp-alist`, which would let me navigate between failed tests using the regular "previous/next error" functions in Emacs.

The problem is that Hedgehog's output is _awesome_ for human consumption, but not so much for Emacs consumption as it doesn't support multi-line error matchers. Even if it did, I couldn't come up with a reliable way of parsing _file_, _line_, and _column_ information.

Would it make sense to add, somewhere in the output of a failing test, a line like:

```
Test failed at: my/path/to/FailedTest.hs:54:4
```

Or, in case the test threw some exception:

```
Test errored at: my/path/to/FailedTest.hs:54:4
```

That would be much more tractable to parse in Emacs (and I guess in other editors, recalling my fun days with Vim's `errorformat`).

Outstanding questions that I can come up with:

* Where in the existing output, if anywhere, could this be placed?
* Do we have _column_ information? If not, just having the line number is still good, so that you can navigate to the failing assertion's line.
* Should this be hidden by a flag/environment variable? Maybe it clutters the existing nice output.

I'd be happy to put together a PR after we've discussed options.

Thanks!

owickstrom:
Fixes #306.

Adds the location of the failed assertion (taken from the `Span` value of the failure report). Example output:

```
 ✗ <interactive> failed at test/Test/Hedgehog/Applicative.hs:91:11
    after 1 test.

       ┏━━ test/Test/Hedgehog/Applicative.hs ━━━
    69 ┃ ...

```

This enables editors to more easily parse the location of failed test assertions, and provide links/jump functionality. This has been tested with Emacs.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -73,7 +73,7 @@ import           System.Directory (makeRelativeToCurrentDirectory)
 import           System.IO (hSetEncoding, stdout, stderr, utf8)
 #endif
 
-import           Text.PrettyPrint.Annotated.WL (Doc, (<+>))
+import           Text.PrettyPrint.Annotated.WL (Doc, (<#>), (<+>))
 import qualified Text.PrettyPrint.Annotated.WL as WL
 import           Text.Printf (printf)
 
@@ -714,7 +714,8 @@ ppProgress name (Report tests discards coverage status) =
     Shrinking failure ->
       pure . icon ShrinkingIcon '↯' . WL.annotate ShrinkingHeader $
         ppName name <+>
-        "failed after" <+>
+        "failed" <+> ppFailedAtLocation (failureLocation failure) <#>
+        "after" <+>
         ppTestCount tests <>
         ppShrinkDiscard (failureShrinks failure) discards <+>
         "(shrinking)"
@@ -725,9 +726,10 @@ ppResult name (Report tests discards coverage result) = do
     Failed failure -> do
       pfailure <- ppFailureReport name tests failure
       pure . WL.vsep $ [
-          icon FailedIcon '✗' . WL.annotate FailedText $
+          icon FailedIcon '✗' . WL.align . WL.annotate FailedText $
             ppName name <+>
-            "failed after" <+>
+            "failed" <+> ppFailedAtLocation (failureLocation failure) <#>
+            "after" <+>
             ppTestCount tests <>
             ppShrinkDiscard (failureShrinks failure) discards <>
             "."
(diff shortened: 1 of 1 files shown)


jacobstanley:
Looking great! :tada: 

I'm surprised this prints like this (extra space shown as `_`):
```
 ✗ <interactive> failed at test/Test/Hedgehog/Applicative.hs:91:11
   _after 1 test.
```

The code looks like that shouldn't be the case, just copy/paste error?

owickstrom:
Hmm, that is strange. I know the `(<#>)` operator is supposed to be replaced with a space when using another operation called `group`, but in this case that shouldn't happen. I'll try with `(<##>)` instead and see how that prints.

hedgehog/src/Hedgehog/Internal/Report.hs
```diff
@@ -757,6 +759,14 @@ ppResult name (Report tests discards coverage result) = do
         ] ++
         ppCoverage tests coverage
 
+ppFailedAtLocation :: Maybe Span -> Doc Markup
+ppFailedAtLocation (Just span') =
+  "at" <+>
+  WL.text (spanFile span') <> ":" <>
+  WL.pretty (unLineNo (spanStartLine span')) <> ":" <>
+  WL.pretty (unColumnNo (spanStartColumn span'))
+ppFailedAtLocation Nothing = mempty
```
jacobstanley: Just a few [style](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md#avoid-pattern-matching-and-guards-in-function-definitions) [nits](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md#avoid-using-primes-when-naming-things), would probably make this:

```haskell
ppFailedAtLocation = \case
  Just x ->
    "at" <+>
    WL.text (spanFile x) <> ":" <>
    WL.pretty (unLineNo (spanStartLine x)) <> ":" <>
    WL.pretty (unColumnNo (spanStartColumn x))
  Nothing ->
    mempty
```

You could also use `span_` or `span0` but I think `x` is fine/neater because it's clear from usage what it is.

owickstrom: Gotcha, thanks! I'll correct those.

owickstrom:
@jacobstanley OK the indent thing was likely a copy-paste error. Confirmed it's correctly indented in GHCi and in regular test runs through `cabal v2-test`.

PR #313, opened 2019-08-09
Make `interleave` logarithmtic rather than linear

edsko:
The version of `interleave` before this PR was closely modelled on my
blog post; a tiny bit _too_ closely; it wasn't the intention of the blog
post to define "good" shrinkers. I had a footnote for that but should
have made it a bit more prominent. Anyway, in this PR I change
`interleave` from being linear (removing one element at a time) to
logarithmic (trying to remove as big a chunk as possible at a time,
trying the whole list, half the list, etc.), matching the behaviour of
QuickCheck's `shrinkList`. This can have rather dramatic consequences
for the performance of shrinking.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Tree.hs
+++ b/hedgehog/src/Hedgehog/Internal/Tree.hs
@@ -301,11 +301,30 @@ splits = \case
     ([], x, xs) :
     fmap (\(as, b, cs) -> (x : as, b, cs)) (splits xs)
 
-dropOne :: Monad m => [NodeT m a] -> [TreeT m [a]]
-dropOne ts = do
-  (xs, _y, zs) <- splits ts
-  pure . TreeT . pure $
-    interleave (xs ++ zs)
+-- | @removes n@ computes all ways we can remove chunks of size @n@ from a list
+--
+-- Examples
+--
+-- > removes 1 [1..3] == [[2,3],[1,3],[1,2]]
+-- > removes 2 [1..4] == [[3,4],[1,2]]
+-- > removes 2 [1..5] == [[3,4,5],[1,2,5],[1,2,3,4]]
+-- > removes 3 [1..5] == [[4,5],[1,2,3]]
+--
+-- Note that the last chunk we delete might have fewer elements than @n@.
+removes :: forall a. Int -> [a] -> [[a]]
+removes k = \xs -> go xs
+  where
+    go :: [a] -> [[a]]
+    go [] = []
+    go xs = xs2 : map (xs1 ++) (go xs2)
+      where
+        (xs1, xs2) = splitAt k xs
+
+dropSome :: Monad m => [NodeT m a] -> [TreeT m [a]]
+dropSome ts = do
+  n   <- takeWhile (> 0) $ iterate (`div` 2) (length ts)
+  ts' <- removes n ts
+  pure . TreeT . pure $ interleave ts'
 
 shrinkOne :: Monad m => [NodeT m a] -> [TreeT m [a]]
 shrinkOne ts = do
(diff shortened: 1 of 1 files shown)


hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -301,11 +301,30 @@ splits = \case
     ([], x, xs) :
     fmap (\(as, b, cs) -> (x : as, b, cs)) (splits xs)
 
-dropOne :: Monad m => [NodeT m a] -> [TreeT m [a]]
-dropOne ts = do
-  (xs, _y, zs) <- splits ts
-  pure . TreeT . pure $
-    interleave (xs ++ zs)
+-- | @removes n@ computes all ways we can remove chunks of size @n@ from a list
```
ruhatch: I get what this is doing, but not quite sure it's clear from this sentence!

'all the ways we can remove chunks of size n' sounds like you'd have:
```
-- > removes 2 [1..4] = [[3,4],[1,4],[1,2]]
```

edsko: Fair. Suggested rewording?

edsko: Would "Splits a list into chunks of size `n` and returns all possible lists where one of these chunks has been removed" be less confusing?

hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -301,11 +301,30 @@ splits = \case
...
+-- Examples
+--
+-- > removes 1 [1..3] == [[2,3],[1,3],[1,2]]
+-- > removes 2 [1..4] == [[3,4],[1,2]]
+-- > removes 2 [1..5] == [[3,4,5],[1,2,5],[1,2,3,4]]
+-- > removes 3 [1..5] == [[4,5],[1,2,3]]
+--
+-- Note that the last chunk we delete might have fewer elements than @n@.
+removes :: forall a. Int -> [a] -> [[a]]
+removes k = \xs -> go xs
```
edsko: Er, this should be eta reduced of course; this is a remnant of an earlier version.

hedgehog/src/Hedgehog/Internal/Tree.hs
jacobstanley: Again, I don't mind the behavior, but I can't reconcile how chunks of size `2` are removed from:
```-- > removes 2 [1..5] == [[3,4,5],[1,2,5],[1,2,3,4]]```

Ah ok, that is this part yes?
> Note that the last chunk we delete might have fewer elements than @n@.

Ok, I think I get it now, posting this ramble for posterity

Comment (edsko) [reply hedgehog/src/Hedgehog/Internal/Tree.hs]: Yeah, it removed the chunk `5 : nothingMoreToRemove` :)

File: hedgehog/src/Hedgehog/Internal/Tree.hs [hunk 3]
```diff
@@ -301,11 +301,30 @@ splits = \case
...
+    go [] = []
+    go xs = xs2 : map (xs1 ++) (go xs2)
+      where
+        (xs1, xs2) = splitAt k xs
+
+dropSome :: Monad m => [NodeT m a] -> [TreeT m [a]]
+dropSome ts = do
+  n   <- takeWhile (> 0) $ iterate (`div` 2) (length ts)
+  ts' <- removes n ts
+  pure . TreeT . pure $ interleave ts'
```
treeowl: There's quite a lot going on here. It's not terribly obvious to me what the overall goal is and how each piece fits into it. Why chunk *this way*? Why choose *these* chunk sizes? How does the recursive call to `interleave` contribute to the whole job? Is there a way to track length information to avoid calling `length` repeatedly?

hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -320,7 +339,7 @@ interleave :: forall m a. Monad m => [NodeT m a] -> NodeT m [a]
 interleave ts =
```
treeowl: `interleave` is begging for documentation. Detailed documentation would be great; barring that, a clear statement of its general goals would be helpful.

PR #323, opened 2019-09-14
Only build-depend on fail and semigroups with ancient GHC

clinty:


jacobstanley:
Is this causing you a problem? `fail` and `semigroups` do that switch internally to make sure the right version is exported afaik

clinty:
We don't have empty packages in Debian so any uses of them (fail, nats, bytestring-builder, etc.) need to be patched out which leads to a lot of wasted labor that needs to be repeated on every package upgrade. Things like semigroups are okay because they're not completely empty, but since you're not actually using `Data.Semigroup.Generic` there's no real point in including it either.

moodmosaic:
>We don't have empty packages in Debian so any uses of them (fail, nats, bytestring-builder, etc.) need to be patched out

What's the current procedure for working around this issue? I don't have Debian around to try this, but both builds are green (just had to re-run the AppVeyor one).

---

If we decide to merge this, at least the dependencies should be aligned with all the others :mag_right:, like so:

image

image

clinty:
Here's an example of how we work around `fail` in the `repline` package: https://sources.debian.org/src/haskell-repline/0.2.1.0-2/debian/patches/no-fail/

(cf. https://github.com/sdiehl/repline/issues/14)

sol:
@moodmosaic from what I see in the current version of the Cabal file, I think this can be closed.

moodmosaic:
Thank you, @sol 🙌 Closing this, though @clinty absolutely feel free to re-open if needed. Thank you for your work on this. 👍

PR #328, opened 2019-09-25
Add a MonadBaseControl instance for PropertyT

treeowl:
I believe it is possible to write a lawful `MonadBaseControl` instance
for `CofreeT` (see [this proof](https://github.com/ekmett/free/issues/193#issuecomment-535173061)), and therefore also for the `TreeT`, `GenT`, and
`PropertyT` types in this package. I can't say for sure whether it
"makes sense", but hey, maybe it does.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -18,6 +18,9 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-} -- MonadBase
+#if __GLASGOW_HASKELL__ >= 806
+{-# LANGUAGE DerivingVia #-}
+#endif
 module Hedgehog.Internal.Gen (
   -- * Transformer
     Gen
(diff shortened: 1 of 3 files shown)


jacobstanley:
Wow this is awesome, I have wanted this for a long time

jacobstanley:
I can't say I really understand this one way or another, but given it won't break any existing code I say we give it a try. Not sure what would be the best way to check this works.

treeowl:
Thanks! Like I said, I really don't have any clear idea of whether the instance is "sensible". The `MonadBaseControl` laws smell rather incomplete to me. Here's hoping!

treeowl:
Thanks! As I said, I don't really know if this instance is reasonable—the `MonadBaseControl` laws smell somewhat incomplete to me, so I don't know just what to look for in that regard. But here's hoping!

jacobstanley:
I've just been trying this out and it does appear to work, at least for the case of using `bracket_` from `lifted-base`. Time will tell I suppose, but I am about to recommend that people use `lifted-base` for bracketing from now on given `ResourceT` has gone the `UnliftIO` route which we'll never be able to implement.

This has saved us from being stuck on an old `ResourceT` forever! :tada: :pray: thanks again

ocharles:
Now that we have `MonadBaseControl PropertyT` is there any need to have `TestT`?

jacobstanley:
That's a good question, it would be interesting to see what happens with `lifted-async` !

This was the original reason for `TestT` if I recall:
https://github.com/hedgehogqa/haskell-hedgehog/blob/master/hedgehog/src/Hedgehog/Internal/State.hs#L819-L837

It would be nice to see if the `linearize` + parallel testing functionality could be made to work with the `lookupIx` style stateful tests.

PR #330, opened 2019-10-09
MonadGen instances for StateT

### LINKED ISSUE #287: Stateful generation
I would like to implement something like `cycle`
```
cycle :: [Gen a] -> Gen a
```
So that each time we generate a value `a` we use a different generator in the original list in a deterministic fashion, one after the other. 

I tried to implement this with something like `GenT (StateT s) a` but couldn't make it work.

Similarly with some stateful generation we could implement something like `distinct :: Gen a -> Gen a` which would keep a list of already generated values and only generate new ones.

Is this possible?

symbiont-sam-halliday:
Assuming the instances work as expected, this **should** fix #287 

I'm fairly sure this is all possible thanks to #321 by @HuwCampbell.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -181,6 +181,8 @@ import           Control.Monad.Trans.Identity (IdentityT(..))
 import           Control.Monad.Trans.Maybe (MaybeT(..))
 import           Control.Monad.Trans.Reader (ReaderT(..))
 import           Control.Monad.Trans.Resource (MonadResource(..))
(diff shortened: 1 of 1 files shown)


sshine:
You're a hero, @symbiont-sam-halliday!

symbiont-sam-halliday:
@jacobstanley could you please release a 1.0.2 with this fix in it? :smile: ... it makes everything so much nicer!

Thanks @sshine but the real hero is @HuwCampbell

symbiont-sam-halliday:
BTW although this is compiling I am seeing some very strange behaviour... it looks like shrunk tests can see state that was created in the unshrunk tests. I had assumed that everything would be correctly isolated but that is not true. This makes me suspicious of all "effectful generators".

PR #338, opened 2019-11-11
Fix up frequency

### LINKED ISSUE #337: frequency is inefficient and wrong
### Correctness

`frequency` is a little bit wrong because it relies on the sum of the frequencies being representable by an `Int`, but does not document this. So `frequency [(maxBound `quot` 2, x), ((maxBound `quot` 4) * 3, y), (2034, z)]` will do something hard to predict. The best fix is likely to use double-word integers in the calculations:

```haskell
data DW = DW {hi :: !Word, lo :: !Word}
```

This is almost certainly available in a library somewhere, but it wouldn't be too hard to roll our own if necessary.

### Efficiency

The argument list is traversed for every value generated. When the list is long, that's really lousy. The solution is to build a tree representing the input list, making lookup much faster. A simple hack using `DW` above would use an `IntMap` of `IntMap`s and some lookup functions like `lookupLE`, `lookupGE`, etc. There are almost certainly better specialized data structures available for the purpose, but I don't know if they'd be worth the trouble.

treeowl:
* Make `frequency` detect negative frequencies and frequency
  total overflow.

* Rewrite `frequency` to build an `IntMap` to represent the
  frequency list, which greatly improves efficiency for long
  lists.

Fixes #337.


treeowl:
I could use another pair of eyes to make sure this produces precisely the right distribution. I believe it does, but I don't trust myself around fence-posts.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1170,28 +1174,38 @@ choice = \case
...
+  do
     n <- integral $ Range.constant 1 total
-    pick n xs0
+    case IM.lookupGE n sum_map of
+      Just (_, a) -> a
+      Nothing -> error "Hedgehog.Gen.frequency: Something went wrong."
+    where
+      --[(1, x), (7, y), (10, z)]  In
+      --[(1, x), (8, y), (18, z)]  Out
+      sum_map = IM.fromDistinctAscList $ List.unfoldr go (0, xs0)
```
treeowl: This probably shouldn't use `unfoldr`. Maybe `mapAccumL`? In particular, I don't think we get any list fusion with `fromDistinctAscList`, but we could get some on the other side. It doesn't actually matter terribly much in practice.

treeowl:
Question: if no frequencies are positive, should we error out (as presently) or should we just discard in that case?

treeowl:
I added a commit that discards when all the frequencies are zero. That feels more correct to me, but feel free to go with whichever you prefer.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1170,28 +1174,38 @@ choice = \case
...
+        where
+          go (_, []) = Nothing
+          go (n, (k, x) : xs)
+            | k < 0 = error "Hedgehog.Gen.frequency: Negative frequency."
+              -- nk < 0 means the sum overflowed.
+            | nk < 0 = error "Hedgehog.Gen.frequency: Frequency sum above maxBound :: Int"
+            | k > 0 = Just ((nk, x), (nk, xs))
+            | otherwise = go (n, xs)
+            where !nk = n + fromIntegral k
```
treeowl: The `fromIntegral` isn't needed. Let's clean that up in merging.

treeowl:
Ah, I just realized we can do a tiny bit better! We can make the `IntMap` one entry *smaller* than the frequency table, holding one element in reserve to handle failed lookup. That removes the "something went wrong" error that's been bothering me.

treeowl:
While the `IntMap` is far better for very long lists, it's definitely not so great for the more typical short lists. What would be good for short lists? Sort the list by frequency in descending order, store the accumulated frequencies in an unboxed vector, store the values in a boxed vector, and do a linear scan.

In some cases, a hybrid mechanism may be best. If a few entries dominate, then it would make sense to store them in vectors, checked first, and the rest in an `IntMap`, as a fall-back. Sounds fussy.

treeowl:
Ping? I haven't seen any comments on this yet.

treeowl:
Hmm.... Is this actually a good idea? I don't know. It only makes sense when we sample multiple times from the generator. That will surely be the case at the "top level", but is it elsewhere? I don't know enough about hedgehog to say. It'd be good to hear from someone who knows something.

treeowl:
Deleted misplaced comment.

HuwCampbell:
Personally I'm not convinced about this one.

Almost always the input to frequency is hand written with a small number of choices and the time will be quite acceptable. It'll probably actually be faster this for most idiomatic use cases.

PR #339, opened 2019-11-12
Stop using fromJust

### LINKED ISSUE #301: filterMaybeT from Hedgehog.Internal.Tree broken
I don't have an easy set of steps to repro this, as it's occurred on a local branch of input-output-hk/cardano-ledger-specs, but I ran into an issue where `runDiscardEffect` was throwing `fromJust: Nothing`.

I hunted this down to `filterMaybeT` in `Hedgehog.Internal.Tree` and found that replacing that with `filterT` from further down in the module fixed the error. From a quick look this could be due to a missing `p x` check on the root of the `Tree` in `filterMaybeT`.

I managed to find a workaround by using `runDiscardEffectT` instead, hence not providing a fix for this. I believe a careful review of that function, and some tests, should hopefully solve this problem.

treeowl:
Get rid of all but one use of `fromJust`.

Fixes #301


treeowl:
Here's my best guess about what this stuff is supposed to do. Please review carefully.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -256,8 +259,7 @@ runGenT size seed (GenT m) =
 --
 evalGen :: Size -> Seed -> Gen a -> Maybe (Tree a)
 evalGen size seed =
-  fmap (fmap Maybe.fromJust) .
-  Tree.filter Maybe.isJust .
+  Tree.mapMaybe id .
```
treeowl: This is a substantial change. Please be especially careful to check that it makes sense!

hedgehog/hedgehog.cabal
```diff
@@ -112,6 +112,11 @@ library
     Hedgehog.Internal.Tree
     Hedgehog.Internal.Tripping
 
+  other-modules:
+    -- This has nothing to do with Hedgehog proper, and it's
+    -- too trivial to bother exposing.
+    Hedgehog.Internal.Predicate
```
jacobstanley: I agree with the comment, but would like to maintain a policy of always exposing all internal modules

hedgehog/src/Hedgehog/Internal/Gen.hs
jacobstanley: Provided `Tree.mapMaybe` works then this looks like it would be fine.

jacobstanley:
I added a stupid unit test just to mess around with it in the repl, looks pretty good to me

jacobstanley:
Just have a few style guide things, will make some suggestions

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
 --   forever. If we trigger these limits then the whole generator is discarded.
 --
 filter :: (MonadGen m, GenBase m ~ Identity) => (a -> Bool) -> m a -> m a
-filter p gen0 =
+filter p = mapMaybe (fromPred p)
```
jacobstanley:
```suggestion
filter p =
  mapMaybe (fromPred p)
```

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
...
         discard
       else do
         (x, gen) <- freeze $ scale (2 * k +) gen0
 
-        if p x then
-          withGenT (mapGenT (Tree.filterMaybeT p)) gen
-        else
-          try (k + 1)
+        case p x of
+          Just _ -> withGenT (mapGenT (Tree.mapMaybeMaybeT p)) gen
```
jacobstanley:
```suggestion
          Just _ ->
            withGenT (mapGenT (Tree.mapMaybeMaybeT p)) gen
```

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
...
+          Nothing -> try (k + 1)
```
jacobstanley:
```suggestion
          Nothing ->
            try (k + 1)
```

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
...
   in
     try 0
 
 filterT :: MonadGen m => (a -> Bool) -> m a -> m a
-filterT p gen0 =
+filterT p = mapMaybeT (fromPred p)
```
jacobstanley:
```suggestion
filterT p =
  mapMaybeT (fromPred p)
```

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
...
         discard
       else do
         (x, gen) <- freeze $ scale (2 * k +) gen0
 
-        if p x then
-          withGenT (mapGenT (Tree.filterT p)) gen
-        else
-          try (k + 1)
+        case p x of
+          Just _ -> withGenT (mapGenT (Tree.mapMaybeT p)) gen
```
jacobstanley:
```suggestion
          Just _ ->
            withGenT (mapGenT (Tree.mapMaybeT p)) gen
```

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1270,34 +1271,38 @@ ensure p gen = do
...
+          Nothing -> try (k + 1)
```
jacobstanley:
```suggestion
          Nothing ->
            try (k + 1)
```

hedgehog/src/Hedgehog/Internal/Predicate.hs
```diff
@@ -0,0 +1,12 @@
+module Hedgehog.Internal.Predicate
+  ( fromPred
+  ) where
+
+import Control.Monad (guard)
+
+-- | Convert a Boolean predicate to something suitable for use
+-- with @mapMaybe@-like functions. The @wizards@ package calls this
+-- @ensure@. @protolude@ and @relude@ generalize from 'Maybe' to
+-- arbitrary 'Control.Applicative.Alternative' and call it @guarded@.
```
moodmosaic: (Extreme nitpick) - need to push those comments on line (8-10) two space deeper, then add an empty `--` line, if we want this to be consistent with all the other commented stuff, like so:

```
-- | Convert a Boolean predicate to something suitable for use
--   with @mapMaybe@-like functions. The @wizards@ package calls this
--   @ensure@. @protolude@ and @relude@ generalize from 'Maybe' to
--   arbitrary 'Control.Applicative.Alternative' and call it @guarded@.
--
```

(Absolutely *not* a blocking issue, of course!) :incoming_envelope:

hedgehog/src/Hedgehog/Internal/Predicate.hs
```diff
@@ -0,0 +1,12 @@
+module Hedgehog.Internal.Predicate
+  ( fromPred
+  ) where
```
jacobstanley: can't do multiline suggestions but
```diff
- module Hedgehog.Internal.Predicate
-   ( fromPred
+ module Hedgehog.Internal.Predicate (
+     fromPred
    ) where
```

hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -249,44 +256,52 @@ runTreeMaybeT =
...
-          []
-        Just x ->
-          [Tree (Node x mxs)]
-    else
-      mxs
+    case mx of
+      Nothing -> mxs
+      Just x ->
+        case p x of
+          Just x' -> [Tree (Node x' mxs)]
```
jacobstanley:
```suggestion
          Just x' ->
            [Tree (Node x' mxs)]
```

hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -249,44 +256,52 @@ runTreeMaybeT =
...
+          Nothing -> mxs
```
jacobstanley:
```suggestion
          Nothing ->
            mxs
```

hedgehog/src/Hedgehog/Internal/Tree.hs
```diff
@@ -249,44 +256,52 @@ runTreeMaybeT =
...
 
 -- | Returns a tree containing only elements that match the predicate.
 --
 --   When an element does not match the predicate its node is replaced with
 --   'empty'.
 --
 filterT :: (Monad m, Alternative m) => (a -> Bool) -> TreeT m a -> TreeT m a
-filterT p m =
+filterT p = mapMaybeT (fromPred p)
```
jacobstanley:
```suggestion
filterT p =
  mapMaybeT (fromPred p)
```

jacobstanley:
Sorry for all the [style](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md) nits, appreciate the contribution!

hedgehog/src/Hedgehog/Internal/Predicate.hs
moodmosaic: Not sure if we even want a dedicated module for this, as we can just declare this in both `Gen` and `Tree` (but not expose it). Then, if we need to use it on a *third* module, we can consider [the rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) and move it to its own module.

It's just that Hedgehog.Internal.Predicate looks like a utility module that we don't really need (yet)...

hedgehog/hedgehog.cabal
moodmosaic: https://github.com/hedgehogqa/haskell-hedgehog/pull/339#discussion_r346418133, most likely we don't even need to introduce a new module.

moodmosaic:
@treeowl would you like me or @jacobstanley to apply those few style nits above? Happy to do so if you're busy. After that we should be good to go and merge this :rocket:

treeowl:
Go ahead. I'm sluggish today.

PR #345, opened 2019-11-24
Fix performance issues with color handling

stolyaroleh:
..by not autodetecting `UseColor` on every `updateSummary` if it is not provided.
`detectColor` is expensive because it calls `hSupportsANSI`, which can result in lookup of 3 environment variables and a call to `IODevice.isTerminal`.

This commit replaces several cases of `Maybe UseColor` with `UseColor` and detects/resolves color explicitly in `check`, `recheck` and `checkGroup`.

See [my earlier comment](https://github.com/hedgehogqa/haskell-hedgehog/pull/240#issuecomment-549049191) for performance numbers I collected while testing.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Report.hs
+++ b/hedgehog/src/Hedgehog/Internal/Report.hs
@@ -1069,8 +1069,8 @@ ppSummary summary =
             Nothing
         ]
 
-renderDoc :: MonadIO m => Maybe UseColor -> Doc Markup -> m String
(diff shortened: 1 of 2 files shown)


stolyaroleh:
Related: #240

HuwCampbell:
This looks very reasonable.

jacobstanley:
I think these are all internal modules so maybe not?

moodmosaic:
Yeah, indeed :+1:

Issue #38, opened 2017-04-11
Documentation for how linear range / size parameter works


chris-martin:
I'm puzzled by the behavior of `linear`.

> Construct a range which scales the second bound relative to the size parameter.

The first example in the doc is `bounds 50 $ linear 0 10`. I'd expect you scale 10 by 50 and you get `(0,500)` but instead you get `(0,5)`.

The other examples seem to confirm that there's an implicit factor of `1/100` multiplied in there somewhere.

And then when the size parameter is above 100, it stops scaling entirely and becomes a constant.

```haskell
λ> Range.bounds 5000 $ Range.linear 0 10
(0,10)
```

So what's the concept here? Is the size parameter in general supposed to be thought of as a percentage, and does it never go above 100?

chris-martin:
I also just noticed that the size parameter used by `sample` is 30 - Is there a particular reason for that?

jystic:
This concept is inherited from QuickCheck and I haven't thought much about whether it could be made better. It does make sense to think of the size parameter as a percentage, it ranges from `0 - 99`, so if you run 200 tests it will cycle back to `0`, then ramp up to `99` again.

I think the general idea is that if the test fails with an input which is initially smaller, then shrinking it is easier, so we should start with inputs that are small. Also, smaller inputs tend to run faster, so only generating a few big inputs helps test performance.

I'm not sure any of these are great reasons for having size as a thing. If we had perfect shrinking every time then I think the concept of size could potentially be removed altogether.

chris-martin:
Ah. That makes sense. I always thought size was an underdocumented concept in quickcheck.

Does it make sense to make this explicit and have the `Size` type constrain its value to the valid range?

If it's a percentage, wouldn't it go to 100?

Is there a reason for it to be integral rather than fractional?

jystic:
> Does it make sense to make this explicit and have the `Size` type constrain its value to the valid range?

Yes but it needs some thought to make sure that doesn't break anything.

> If it's a percentage, wouldn't it go to 100?

I guess 0-99 makes more sense as that is exactly 100 levels and the default number of tests is 100.

> Is there a reason for it to be integral rather than fractional?

Only that it matches what QuickCheck does, maybe being 0 - 1 as a fractional would be better, I'm not sure.

moodmosaic:
>This concept is inherited from QuickCheck and I haven't thought much about whether it could be made better.

[...]

PR #346, opened 2019-11-27
Some doc on the bounds of Size (#38)

chris-martin:
There is probably more we can do to address #38, but immediately I think that just mentioning that the bounds of `Size` are 0 and 99 is a good start.


moodmosaic:
Thank you :100:

PR #347, opened 2019-12-24
A few optimizations

treeowl:
* Use arrays to speed up lookups in `element` and `choice`; each
  lookup is now O(1) instead of O(n).

* Use `Data.List.inits` to make `splits` considerably faster
  and more memory efficient.

* Use `Data.Sequence` to make `shuffle` asymptotically more efficient.
  Expose the new `shuffleSeq` function used to implement `shuffle`.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -213,7 +213,7 @@ import qualified Data.Maybe as Maybe
 import qualified Data.Primitive.SmallArray as PSArr
 import           Data.Semigroup (Semigroup)
 import qualified Data.Semigroup as Semigroup
-import           Data.Sequence (Seq)
+import           Data.Sequence (Seq (..))
 import qualified Data.Sequence as Seq
 import           Data.Set (Set)
 import           Data.Text (Text)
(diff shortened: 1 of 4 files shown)


hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1589,16 +1597,22 @@ subsequence xs =
...
-    case splitAt n xs0 of
-      (xs, y : ys) ->
-        (y :) <$> shuffle (xs ++ ys)
-      (_, []) ->
-        error "Hedgehog.Gen.shuffle: internal error, split generated empty list"
+-- We shuffle sequences instead of lists to make extracting an arbitrary
+-- element logarithmic instead of linear, and to make length calculation
+-- constant-time instead of linear. We could probably do better, but at
+-- least this should be less intolerably slow.
+shuffle = fmap toList . go . Seq.fromList
```
treeowl: We should break out `go` into a top-level function and expose it.

treeowl:
Some informal testing indicates that this is indeed *much* faster than the old implementation for non-small lists.

treeowl:
I'm going to split this up and remove part of it.

PR #198, opened 2018-05-23
 Avoid weak gamma values in Hedgehog.Internal.Seed

moodmosaic:
Resolves https://github.com/hedgehogqa/haskell-hedgehog/issues/191 ― See https://github.com/hedgehogqa/haskell-hedgehog/issues/191#issuecomment-391209131 ― ✔ Passes all dieharder tests. :neckbeard:
 
---

```
$ dieharder-input | dieharder -a -g 200
#=============================================================================#
#            dieharder version 3.31.1 Copyright 2003 Robert G. Brown          #
#=============================================================================#
   rng_name    |rands/second|   Seed   |
stdin_input_raw|  3.00e+07  |2424813648|
#=============================================================================#
        test_name   |ntup| tsamples |psamples|  p-value |Assessment
#=============================================================================#
   diehard_birthdays|   0|       100|     100|0.88800885|  PASSED  
      diehard_operm5|   0|   1000000|     100|0.19714083|  PASSED  
  diehard_rank_32x32|   0|     40000|     100|0.38815709|  PASSED  
    diehard_rank_6x8|   0|    100000|     100|0.94344462|  PASSED  
   diehard_bitstream|   0|   2097152|     100|0.96453841|  PASSED  
        diehard_opso|   0|   2097152|     100|0.06116614|  PASSED  
        diehard_oqso|   0|   2097152|     100|0.23506030|  PASSED  
         diehard_dna|   0|   2097152|     100|0.40548706|  PASSED  
diehard_count_1s_str|   0|    256000|     100|0.93024945|  PASSED  
diehard_count_1s_byt|   0|    256000|     100|0.37068664|  PASSED  
 diehard_parking_lot|   0|     12000|     100|0.35955403|  PASSED  
    diehard_2dsphere|   2|      8000|     100|0.99480570|  PASSED  
    diehard_3dsphere|   3|      4000|     100|0.90990947|  PASSED  
     diehard_squeeze|   0|    100000|     100|0.07403175|  PASSED  
        diehard_sums|   0|       100|     100|0.00188117|   WEAK   
        diehard_runs|   0|    100000|     100|0.86278014|  PASSED  
        diehard_runs|   0|    100000|     100|0.55399860|  PASSED  
       diehard_craps|   0|    200000|     100|0.72844871|  PASSED  
       diehard_craps|   0|    200000|     100|0.66463648|  PASSED  
 marsaglia_tsang_gcd|   0|  10000000|     100|0.21806049|  PASSED  
[...]

PR #348, opened 2019-12-25
Speed up shuffle

treeowl:
* Add `shuffleSeq` to shuffle sequences.

* Use `shuffleSeq` to implement `shuffle` much more efficiently. The shuffling
  algorithm we use finds and removes an arbitrary element, which sequences
  can do in logarithmic time and space rather than linear time and
  space. It also uses `length`, which is constant time for sequences and
  linear time for lists.

* Bump lower bound on `containers`. We could avoid this, if necessary,
  by not using pattern synonyms for sequences.

### DIFF
--- a/hedgehog/src/Hedgehog/Gen.hs
+++ b/hedgehog/src/Hedgehog/Gen.hs
@@ -90,6 +90,7 @@ module Hedgehog.Gen (
   -- ** Combinations & Permutations
   , subsequence
   , shuffle
+  , shuffleSeq
 
   -- ** Abstract State Machine
   , sequential
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -125,6 +125,7 @@ module Hedgehog.Internal.Gen (
   -- ** Combinations & Permutations
   , subsequence
   , shuffle
+  , shuffleSeq
 
   -- * Sampling Generators
   , sample
@@ -1589,16 +1590,40 @@ subsequence xs =
 --   /list./
 --
 shuffle :: MonadGen m => [a] -> m [a]
-shuffle = \case
-  [] ->
-    pure []
-  xs0 -> do
-    n <- integral $ Range.constant 0 (length xs0 - 1)
-    case splitAt n xs0 of
-      (xs, y : ys) ->
-        (y :) <$> shuffle (xs ++ ys)
-      (_, []) ->
-        error "Hedgehog.Gen.shuffle: internal error, split generated empty list"
+-- We shuffle sequences instead of lists to make extracting an arbitrary
+-- element logarithmic instead of linear, and to make length calculation
+-- constant-time instead of linear. We could probably do better, but
+-- this is at least reasonably quick.
+shuffle = fmap toList . shuffleSeq . Seq.fromList
+
+-- | Generates a random permutation of a sequence.
+--
+--   /This shrinks towards the order of the sequence being identical to the input/
+--   /sequence./
+--
+shuffleSeq :: MonadGen m => Seq a -> m (Seq a)
+shuffleSeq xs
+  | null xs = pure Seq.empty
+  | otherwise = do
+    n <- integral $ Range.constant 0 (length xs - 1)
+#if MIN_VERSION_containers(0,5,8)
+    -- Data.Sequence should offer a version of deleteAt that returns the
+    -- deleted element, but it does not currently do so. Lookup followed
+    -- by deletion seems likely faster than splitting and then appending,
+    -- but I haven't actually tested that. It's certainly easier to see
+    -- what's going on.
+    case Seq.lookup n xs of
+      Just y -> (y Seq.<|) <$> shuffleSeq (Seq.deleteAt n xs)
+      Nothing -> error $
+        "Hedgehog.Gen.shuffleSeq: internal error, lookup in empty sequence"
+#else
+    case Seq.splitAt n xs of
+      (beginning, end) ->
+        case Seq.viewl end of
+          y Seq.:< end' -> (y Seq.<|) <$> shuffleSeq (beginning Seq.>< end')
+          Seq.EmptyL -> error $
+            "Hedgehog.Gen.shuffleSeq: internal error, lookup in empty sequence"
+#endif
 
 ------------------------------------------------------------------------
 -- Sampling


HuwCampbell:
This looks good to me; some benchmarking results would be good though.

With regards to bumping containers, I don't think it should be necessary as it looks like the pattern synonym is only used for adding an element, so plain `(<|)` should suffice.

treeowl:
I used it for more earlier, but shifted. Yeah, I'll take that out.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1589,16 +1590,28 @@ subsequence xs =
...
+shuffleSeq Empty = pure Empty
+shuffleSeq xs0 = do
+  n <- integral $ Range.constant 0 (length xs0 - 1)
+  -- Data.Sequence should offer a version of deleteAt that returns the
+  -- deleted element, but it does not currently do so. Lookup followed
+  -- by deletion seems likely faster than splitting and then appending.
+  case Seq.lookup n xs0 of
+    Just y -> (y :<|) <$> shuffleSeq (Seq.deleteAt n xs0)
+    Nothing ->
+      error "Hedgehog.Gen.shuffleSeq: internal error--lookup in empty sequence"
```
moodmosaic: `"Hedgehog.Gen.shuffleSeq: internal error--lookup in empty sequence"` vs
`"Hedgehog.Gen.shuffleSeq: internal error, lookup in empty sequence"`

I think the second one is grammatically better.

moodmosaic:
@treeowl, thank you, for all the hard work you've been putting in here and on the other PRs :+1: :beers: Often we're all swamped and so there're delays. (Note to self, perhaps it's time for a `CONTRIBUTING.md`.)

---

Agreed with @HuwCampbell on trying to avoid bumping dependencies when possible.

Also, it'd be nice with some benchmarks for future reference, as a comment here, or better in the commit message itself. (Yep, we've had some fairly [long commit messages](https://github.com/hedgehogqa/haskell-hedgehog/pull/198/commits/39b15b9b4d147f6001984c4b7edab00878269da7) in the past.)

Once we resolve those, we could look forward getting this merged /cc @jacobstanley

treeowl:
I've done everything except benchmarking. I don't really have the time right now to set up Criterion benchmarks for this. Messing around in GHCi with moderately long lists seemed a pretty convincing demonstration, and it drops the algorithm from quadratic to linearithmic in both time and allocation.

treeowl:
@moodmosaic any chance you could have another look at this?

moodmosaic:
Yep, during the weekend hopefully, as soon as I'm back at my laptop.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1589,16 +1590,40 @@ subsequence xs =
...
+    -- Data.Sequence should offer a version of deleteAt that returns the
+    -- deleted element, but it does not currently do so. Lookup followed
+    -- by deletion seems likely faster than splitting and then appending,
+    -- but I haven't actually tested that. It's certainly easier to see
+    -- what's going on.
+    case Seq.lookup n xs of
+      Just y -> (y Seq.<|) <$> shuffleSeq (Seq.deleteAt n xs)
+      Nothing -> error $
+        "Hedgehog.Gen.shuffleSeq: internal error, lookup in empty sequence"
+#else
```
moodmosaic: image

The code in `#if MIN_VERSION_containers(0,5,8)` (shown as `Gen.shuffle_fast_1`) is actually faster than the one in the `#else` branch (shown as `Gen.shuffle_fast_2`.)

Do we need to have both? It would be good to avoid CPP where possible.

moodmosaic: Oh, is it because of `containers                      >= 0.4        && < 0.7`?

treeowl:
Yeah, `deleteAt` is relatively new. I'm glad my guess about the relative performance of those turned out right.

moodmosaic:
Yep, FTR it was added in 0.5.2 https://hackage.haskell.org/package/containers-0.5.2.0/docs/doc-index-D.html

But since the other one in the `#else` branch is nearly as fast, perhaps we could just use that one. We could add a `FIXME` there with a link pointing at this discussion and update it once we bump `containers`.

Just to avoid the extra maintenance and the CPP is all, but absolutely open for keeping both if you there's a valid reason for it.

moodmosaic:
OTOH the one in the `#else` branch is ugly (and a bit slower) compared to the one using `deletedAt`

moodmosaic:
`containers` v0.4 is 10 years old, could very well be a reason to bump the lower version to v0.5.2...

treeowl:
Only you can decide what bounds you consider acceptable. I think it would be a shame not to take the improved performance when we can get it. Do you think you could break up the benchmarks a bit? It would be useful to see scaled comparisons (or just text) for 3/30/300 and also 3000 elements. The small end in the graph above is too small to see, and at the large end you don't seem to get quite large enough for the new algorithm to blow the old one out of the water.

moodmosaic:
👍 Sounds reasonable, will do and try to upload the full benchmark HTML here.

moodmosaic:
The one taking advantage of `containers` 0.5.8 is nearly twice faster than the one that works with 0.4.0 :nerd_face: 

image

```
benchmarking current/3
time                 2.186 μs   (2.084 μs .. 2.326 μs)
                     0.982 R²   (0.978 R² .. 0.988 R²)
mean                 2.301 μs   (2.209 μs .. 2.395 μs)
std dev              324.8 ns   (271.9 ns .. 399.3 ns)
variance introduced by outliers: 94% (severely inflated)

benchmarking current/30
time                 26.13 μs   (25.69 μs .. 26.59 μs)
                     0.995 R²   (0.992 R² .. 0.997 R²)
mean                 28.04 μs   (27.23 μs .. 29.02 μs)
std dev              2.849 μs   (2.200 μs .. 3.849 μs)
variance introduced by outliers: 85% (severely inflated)

benchmarking current/300
time                 2.187 ms   (2.103 ms .. 2.283 ms)
                     0.988 R²   (0.982 R² .. 0.994 R²)
mean                 2.122 ms   (2.078 ms .. 2.179 ms)
std dev              173.2 μs   (139.5 μs .. 215.5 μs)
variance introduced by outliers: 59% (severely inflated)

benchmarking current/3000
time                 184.7 ms   (177.6 ms .. 191.6 ms)
                     0.999 R²   (0.997 R² .. 1.000 R²)
mean                 197.0 ms   (190.4 ms .. 215.1 ms)
std dev              14.72 ms   (435.9 μs .. 20.86 ms)
variance introduced by outliers: 15% (moderately inflated)

benchmarking faster_min_containers_0.5.8/3
time                 2.108 μs   (2.050 μs .. 2.189 μs)
                     0.992 R²   (0.986 R² .. 0.997 R²)
mean                 2.155 μs   (2.112 μs .. 2.212 μs)
std dev              170.6 ns   (135.0 ns .. 225.7 ns)
variance introduced by outliers: 82% (severely inflated)

benchmarking faster_min_containers_0.5.8/30
time                 28.31 μs   (26.20 μs .. 29.86 μs)
                     0.979 R²   (0.970 R² .. 0.989 R²)
mean                 26.60 μs   (25.68 μs .. 27.72 μs)
std dev              3.244 μs   (2.673 μs .. 3.972 μs)
variance introduced by outliers: 89% (severely inflated)

benchmarking faster_min_containers_0.5.8/300
time                 405.3 μs   (381.2 μs .. 423.8 μs)
                     0.974 R²   (0.963 R² .. 0.983 R²)
mean                 439.3 μs   (417.8 μs .. 464.5 μs)
std dev              80.82 μs   (66.11 μs .. 99.06 μs)
variance introduced by outliers: 93% (severely inflated)

benchmarking faster_min_containers_0.5.8/3000
time                 9.775 ms   (9.378 ms .. 10.33 ms)
                     0.978 R²   (0.955 R² .. 0.993 R²)
mean                 11.50 ms   (11.04 ms .. 12.18 ms)
std dev              1.599 ms   (1.160 ms .. 2.401 ms)
variance introduced by outliers: 68% (severely inflated)

benchmarking faster_min_containers_0.4.0/3
time                 2.435 μs   (2.310 μs .. 2.550 μs)
                     0.988 R²   (0.980 R² .. 0.995 R²)
mean                 2.270 μs   (2.224 μs .. 2.345 μs)
std dev              194.2 ns   (138.1 ns .. 304.8 ns)
variance introduced by outliers: 84% (severely inflated)

benchmarking faster_min_containers_0.4.0/30
time                 32.13 μs   (31.03 μs .. 33.06 μs)
                     0.991 R²   (0.985 R² .. 0.994 R²)
mean                 30.44 μs   (29.48 μs .. 31.42 μs)
std dev              3.372 μs   (2.949 μs .. 3.863 μs)
variance introduced by outliers: 87% (severely inflated)

benchmarking faster_min_containers_0.4.0/300
time                 805.6 μs   (750.3 μs .. 863.3 μs)
                     0.967 R²   (0.950 R² .. 0.983 R²)
mean                 733.4 μs   (696.7 μs .. 771.1 μs)
std dev              121.6 μs   (100.9 μs .. 152.5 μs)
variance introduced by outliers: 89% (severely inflated)

benchmarking faster_min_containers_0.4.0/3000
time                 16.59 ms   (15.38 ms .. 17.80 ms)
                     0.978 R²   (0.960 R² .. 0.991 R²)
mean                 20.03 ms   (19.02 ms .. 21.93 ms)
std dev              3.180 ms   (1.917 ms .. 5.463 ms)
variance introduced by outliers: 71% (severely inflated)
```

hedgehog/src/Hedgehog/Gen.hs
```diff
@@ -90,6 +90,7 @@ module Hedgehog.Gen (
   -- ** Combinations & Permutations
   , subsequence
   , shuffle
+  , shuffleSeq
```
moodmosaic: `shuffleSeq` can be part of the public `Hedgehog.Gen` API later if needed, e.g. if we get many use cases or requests for it. Because once we add it there, it's basically there forever.

For now let's keep it as part of the `Hedgehog.Internal.Gen` API as you've added it below. Apart from that, we can merge this and fine tune it later (e.g. remove the `#else` CPP branch as per the bench results shown earlier on).

moodmosaic:
```suggestion
```

PR #356, opened 2020-01-26
Add example using lifted-base/bracket for resource handling

jacobstanley:


massysett:
Thanks for this - I was puzzling over how to do this tonight and then this solved it perfectly; was glad to see it was added just 20 days ago.

PR #359, opened 2020-01-26
Import hedgehog-corpus

jacobstanley:
Moving this from https://github.com/tmcgilchrist/hedgehog-corpus

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,5 +1,6 @@
 packages:
   hedgehog
+  hedgehog-corpus
   hedgehog-dieharder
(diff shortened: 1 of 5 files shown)


HuwCampbell:
👍 on the idea.

It might be reasonable to import the repo into a directory of the source tree and do a PR to merge the tree into the current one.

jacobstanley:
I figured the history wasn't that valuable, in reality most are from Ambiata anyway

HuwCampbell:
Fair.

PR #365, opened 2020-01-30
Bump pretty-show to 1.10 which supports quasi-quotes

jacobstanley:
I have bumped the bounds and updated the hedgehog-example:
```
       ┏━━ hedgehog-example/src/Test/Example/QuasiShow.hs ━━━
    35 ┃ prop_foo :: Property
    36 ┃ prop_foo =
    37 ┃   property $ do
    38 ┃     x <- forAll (SomeData <$> genFoo <*> genFoo)
       ┃     │ SomeData { someFoo1 = [foo|kick|] , someFoo2 = [foo|kick|] }
    39 ┃     x === SomeData (Foo "quasi") (Foo "quoted")
       ┃     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       ┃     │ ━━━ Failed (- lhs) (+ rhs) ━━━
       ┃     │ - SomeData { someFoo1 = [foo|kick|] , someFoo2 = [foo|kick|] }
       ┃     │ + SomeData { someFoo1 = [foo|quasi|] , someFoo2 = [foo|quoted|] }
```

I updated the bounds on Hackage for `hedgehog-1.0.2` already.

This will be particularly useful for types like dates and times which have a sensible human readable representation.

By changing the `Show` instance to render with quasi-quotes you can stay within the unwritten rule that `Show` should be valid Haskell syntax to construct the value, while still having something which is very compact.

For example:
```
[time|2000-01-01T10:00:00|]
```
vs
```
DateTime { datetimeDate = Date { dateYear = Year { getYear = 2000 }, ... etc
```

The means you should be able to both read counterexamples easily and also paste them into your code for experimentation.

Related: https://github.com/yav/pretty-show/issues/40

### DIFF
--- a/hedgehog-example/hedgehog-example.cabal
+++ b/hedgehog-example/hedgehog-example.cabal
@@ -39,6 +39,7 @@ library
     , Test.Example.Coverage
     , Test.Example.Exception
     , Test.Example.List
+    , Test.Example.QuasiShow
     , Test.Example.QuickCheck
     , Test.Example.References
     , Test.Example.Registry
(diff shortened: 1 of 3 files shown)


moodmosaic:
:rocket:

PR #366, opened 2020-02-01
Add filterT, justT, and mapMaybeT to Gen exports.

kquick:
Allows the functionality these represent to be used when the GenBase m
is not Identity.


jacobstanley:
Thanks, looks like an oversight! :tada:

PR #374, opened 2020-02-22
tripping with a monadic action

ersran9:
adds a function `trippingM` which can do `tripping` with a monadic action


ersran9:
I'm not sure why appveyor 8.6.5 and 8.8 is failing... anything we can do to fix it?

moodmosaic:
>I'm not sure why appveyor 8.6.5 and 8.8 is failing

@ersran9, sometimes it's just some transient error, so I've just manually triggered another build. Let's see how that goes.

PR #380, opened 2020-05-06
Add parallel' and sequential' that take a generator of commands

markus1189:
More general than the unticked version that only takes a list and uses
`Gen.element` to choose one.

This allows you to, for example, use `Gen.frequency` to generate commands with different weights.


markus1189:
Does this have a chance of being merged?

ocharles:
This looks like a good addition to me. Hopefully we can get it in in some form (but I'm not a maintainer, so I won't be able to help there).

Issue #245, opened 2019-01-14
`either` combinator?


chessai:
Just found myself wanting this:

```haskell
either :: MonadGen m => m e -> m a -> m (Either e a)
either genE genA =
  sized $ \n ->
    frequency [
        (2, Left <$> genE)
      , (1 + fromIntegral n, Right <$> genA)
      ]
```

moodmosaic:
Resolved by #382.

Thank you both @dcastro and @chessai.

PR #382, opened 2020-05-16
Add either and either_

dcastro:
This PR adds the `either` generator suggested by @chessai in #245, and also an unbiased `either_` version.

Regarding documentation: I grouped these generators in the same section as the `maybe` generator, but I'm not too sure about it being under "Collections". What do the maintainers think?


moodmosaic:
:tada:

PR #384, opened 2020-05-30
Add evalNF

dcastro:
Added `evalNF` as discussed in #383.

A couple of things:

1. I created a local `failException'`, which behaves exactly like `failException` except it also prints `━━━ Value could not be evaluated to normal form ━━━`. Alternatively, we could simply change the type of `failException` to take an additional list of lines to be printed.
    ```haskell
    -- before
    failException :: (MonadTest m, HasCallStack) => SomeException -> m a

    -- after
    failException :: (MonadTest m, HasCallStack) => [String] -> SomeException -> m a
   ```
   I suppose since `failException` is internal this would not be a breaking change, but still inconvenient... Any thoughts on whether we should refactor `failException` or leave the local bind for `failException'` in?
2. It wasn't clear to me how to add new tests for `evalNF`. I tried searching for tests for the existing `eval`, but couldn't find any. Should I add any tests and, if so, to which project/module?
3. I forgot to re-export `either` and `either_` from `Hedgehog.Gen` in my previous PR, so I did that here, if that's OK. I can also submit this is a separate PR if you'd prefer.

### DIFF
--- a/hedgehog-example/hedgehog-example.cabal
+++ b/hedgehog-example/hedgehog-example.cabal
@@ -38,6 +38,7 @@ library
       Test.Example.Basic
     , Test.Example.Confidence
     , Test.Example.Coverage
+    , Test.Example.EvalNF
     , Test.Example.Exception
     , Test.Example.List
     , Test.Example.QuasiShow
--- a/hedgehog-example/src/Test/Example/EvalNF.hs
+++ b/hedgehog-example/src/Test/Example/EvalNF.hs
@@ -1,0 +1,24 @@
+{-# LANGUAGE TemplateHaskell #-}
+module Test.Example.EvalNF where
+
+import           Data.Functor (void)
+
+import           Hedgehog
+
+prop_evalNF_success :: Property
+prop_evalNF_success =
+  property $
+    void $
+      evalNF [""]
+
+prop_evalNF_failure :: Property
+prop_evalNF_failure =
+  property $
+    void $
+      evalNF ["", undefined]
+
+------------------------------------------------------------------------
+
+tests :: IO Bool
+tests =
+  checkSequential $$(discover)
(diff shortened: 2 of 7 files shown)


jacobstanley:
I'm ok with modifying the existing `failException` to take an argument given it's internal :+1:

jacobstanley:
> It wasn't clear to me how to add new tests for evalNF. I tried searching for tests for the existing eval, but couldn't find any. Should I add any tests and, if so, to which project/module?

I think we don't have any tests for the `evalXXX` functions at the moment.

You could add a module to `hedgehog-example` if you're so inclined. Note that the tests in this project don't fail the build, all of the tests are failing on purpose to demonstrate how to use different features.

I'm cool with having no test though as it's so trivial.

moodmosaic:
@jacobstanley @dcastro 
>I'm ok with modifying the existing `failException` to take an argument given it's internal :+1: 

Or even make a new top-level one called `failExceptionWith` taking a message, following the pattern used in all other `xyzWith` ones. Then have `failException` use the `failExceptionWith` one.

moodmosaic:
Yep, basically we use `eval` in some of the [examples](https://github.com/hedgehogqa/haskell-hedgehog/tree/master/hedgehog-example) via the [`assert`](https://github.com/hedgehogqa/haskell-hedgehog/blob/4ec7254df66b5bd95369dfe92dc28dd48f7f5264/hedgehog/src/Hedgehog/Internal/Property.hs#L726-L732) one:

https://github.com/hedgehogqa/haskell-hedgehog/blob/4ec7254df66b5bd95369dfe92dc28dd48f7f5264/hedgehog/src/Hedgehog/Internal/Property.hs#L726-L732

Here's one of those examples:

https://github.com/hedgehogqa/haskell-hedgehog/blob/4ec7254df66b5bd95369dfe92dc28dd48f7f5264/hedgehog-example/src/Test/Example/Basic.hs#L50-L55

dcastro:
Thanks for the feedback! 

I've added `failExceptionWith`, changed `failException` to use `failExceptionWith`, and added a couple of tests to `hedgehog-example`.

moodmosaic:
Thank you, @dcastro!

PR #390, opened 2020-06-26
Bump cabal-version lower bound

moodmosaic:
I'm not sure if there was an announcement for it, but it appears that Hackage now rejects packages with cabal-version lower bound less than 1.10:

```
'cabal-version' must be at least 1.10
```

And because of 1.10 it's also required to add the `default-language` field.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -29,7 +29,7 @@ license:
 license-file:
   LICENSE
 cabal-version:
-  >= 1.8
+  >= 1.10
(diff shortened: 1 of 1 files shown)


HuwCampbell:
Can you explain why default-language is required? I'm doing an optparse release this week, so will need to do this as well.

moodmosaic:
Hackage wouldn't allow me to upload the package candidate otherwise. It was giving me an error that the default-language must be specified...

Also doing a build locally yields the same (as warning though).

PR #391, opened 2020-07-31
Clean up hedgehog.cabal for GHC 8.0+

felixonmars:
Since we have dropped support for GHC 7.10, the redundant packages `fail` and `semigroups` can be removed now. The additional warning is also included in `-Wall`.


moodmosaic (approved):
:+1:

PR #392, opened 2020-08-11
GHC 9 Support

utdemir:
This PR makes Hedgehog compile with the current GHC development branch (soon to be GHC 9). It only required two minor changes on hedgehog side:

* Use `Data.Kind.Type` instead of `*`
* Supporting template-haskell 2.17

See the commit messages for a bit more details.

These changes should be enough from the hedgehog side, but even after these:

* `--allow-newer` flag has to be passed since `primitive` has an overly restrictive upper bound on `base`.
* `QuickCheck` does not compile [yet](https://github.com/nick8325/quickcheck/pull/311), so `hedgehog-quickcheck` does not work yet.

I tried to keep the changes as small as possible while maintaining the backwards compatibility, however I did only check with GHC 8.10 and GHC 9. Let me know if there is anything else I should do to get it compatible with even older GHC versions.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -71,7 +71,7 @@ library
     , resourcet                       >= 1.1        && < 1.3
     , semigroups                      >= 0.16       && < 0.20
     , stm                             >= 2.4        && < 2.6
-    , template-haskell                >= 2.10       && < 2.17
+    , template-haskell                >= 2.10       && < 2.18
     , text                            >= 1.1        && < 1.3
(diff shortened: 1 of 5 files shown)


alexbiehl:
With GHC-9.0.1 released it would be great if this PR could be merge.d

moodmosaic:
Thank you, @utdemir, we apologize for the radio silence. This should be out on the next release (coming soon) :rocket:

moodmosaic (commented):
:+1: /cc co-reviewed @jacobstanley

PR #404, opened 2020-10-07
Add IO capabilities to Ensure

### LINKED ISSUE #402: Change Ensure to be polymorphic over m
Currently we have 

```haskell
Ensure :: (state Concrete -> state Concrete -> input Concrete -> output -> Test ()) -> Callback input output state
```

I propose we change this to

```haskell
Ensure :: (state Concrete -> state Concrete -> input Concrete -> output -> m ()) -> Callback input output state
```

This will also need us to change

```haskell
commandCallbacks :: [Callback input output state]
```

to be

```haskell
commandCallbacks :: [Callback input output m state]
```

My argument is `Command` already has an `m` which in practice will have to be `MonadTest` (as required by `executeSequential`). By using `m` here, we can use all the ambient effects that are available when we execute steps from commands. In particular, this lets me call other parts of my API to make sure they agree.

To provide an example, I'd like to be able to write something like:

```haskell
cmd = Command { commandExecute, commandCallbacks = [ Ensure check ] }
  where
    commandExecute .. = insertNewUser
    check before after username userId = do
      username' <- getUserById userId
      username == username'
```

Right now this is much harder than it should be, because `check` is a `Test ()`, so I'm going to have to give my `Command` some kind of `MyMonad a -> Test a` natural transformation to be able to run `getUserById :: MyApp m => UserId -> m a`.

If the maintainers of Hedgehog agree, I'll prepare a pull request.

ocharles:
Fixes #402


ocharles:
I'm curious what people think about this. In #402 I thought it would maybe be a bit more straight forward, but I see now that `Ensure` is really about pure tests. By adding `m` to `Ensure`, we're really opening ourselves up to "impure" tests - tests that have the ability to consult the very state machine they are testing. Maybe this is not such a good idea

HuwCampbell:
Not much of a comment, but the fact that this means that there will be lots of IO during a search to linearize the function the branches makes this a bit scary.

ocharles:
I'm actually changing my feeling on this too. So far I've just put my "IO Ensure checks" in as separate commands to run the action and then do a pure Ensure. I'll keep this around for more discussion, but I think we may be ok without this one

Issue #118, opened 2017-09-06
Duplicates in shrink trees for maybe


enolan:
The shrink tree for this generator contains several duplicates:
```
$ stack ghci --package hedgehog --resolver lts-9.2
Configuring GHCi with the following packages: 
GHCi, version 8.0.2: http://www.haskell.org/ghc/  :? for help
Loaded GHCi configuration from /tmp/ghci18165/ghci-script
Prelude> import Hedgehog
Prelude Hedgehog> import qualified Hedgehog.Gen as Gen
Prelude Hedgehog Gen> Gen.print $ Gen.maybe Gen.latin1
=== Outcome ===
Just 'T'
=== Shrinks ===
Nothing
Just 'T'
Just 'T'
Just 'T'
Just 'T'
Just '\NUL'
Just '*'
Just '?'
Just 'J'
Just 'O'
Just 'R'
Just 'S'
```

This slows down shrinking a lot. I think the culprit is `maybe`, you get reasonable shrinks from `latin1` on its own.

jystic:
I'm not sure how to avoid this in the generators themselves, but perhaps we can build some things in to the tree to prevent duplicate entries if you're finding that shrinking is slowed down a lot by this.

TysonMN:
@HuwCampbell said in https://github.com/hedgehogqa/haskell-hedgehog/pull/406#issuecomment-709617637 that issue was fixed by that PR.  Does someone want to test if this issue is gone?

moodmosaic:
@TysonMN, here are some sample outputs:

```
λ Gen.print $ Gen.maybe Gen.latin1
=== Outcome ===
Just '@'
=== Shrinks ===
Nothing
Just '\NUL'
Just ' '
Just '0'
Just '8'
Just '<'
Just '>'
Just '?'
```

```
λ Gen.print $ Gen.maybe Gen.latin1
=== Outcome ===
Just 'S'
=== Shrinks ===
Nothing
Just '\NUL'
Just '*'
Just '?'
Just 'I'
Just 'N'
Just 'Q'
Just 'R'
```

It appears that #406 closes this one.

PR #406, opened 2020-10-14
Change frequency generator to immediately shrink

ocharles:
Change frequency generator immediately shrink

The frequency generators works by totalling all weights, and then
picking a random number between [0, total]. It then selects the first
generator who's cummulative weight matches this number. This works fine,
but the current implementation leads to poor shrinking behavior. In
particular, the shrinking is driven by repeatedly picking smaller
numbers in the range [0, total]. For skewed distributions, this can mean
repeatedly reselecting the same generator.

This commit makes the observation that the weight only matters for the
first pick. Once you have made a choice and observed a test failure, you
can give all generators equal weight, as we're going to perform a
breadth-first enumeration to try and find a shrink anyway.

Before this patch, we would see behavior like:

```
> printWith 30 (Seed 4 1) $ Hedgehog.Gen.maybe alphaNum
=== Outcome ===
Just 'o'
=== Shrinks ===
Nothing
Just 'o'
Just 'o'
Just 'o'
Just 'o'
Just 'a'
Just 'h'
Just 'l'
Just 'n'
```

Note that `Just 'o'` appears many times - this is because the maybe
generator puts a heavy bias towards 'Just' constructors.

With this commit, we get the much more minimal output:

```
> printWith 30 (Seed 4 1) $ Hedgehog.Gen.maybe alphaNum
=== Outcome ===
Just 'o'
=== Shrinks ===
Nothing
Just 'a'
Just 'h'
Just 'l'
Just 'n'
```

Note that no characters were duplicated.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -1190,10 +1190,13 @@ frequency = \case
           else
             pick (n - k) xs
 
+      iis =
+        scanl1 (+) (fmap fst xs0)
+
       total =
         sum (fmap fst xs0)
 
-    n <- integral $ Range.constant 1 total
+    n <- shrink (\n -> takeWhile (< n) iis) $ integral_ $ Range.constant 1 total
     pick n xs0
 
 -- | Modifies combinators which choose from a list of generators, like 'choice'


TysonMN:
Why does the outcome change?  (`Just 'o'` vs `Just 'I'`)

ocharles:
Oh, I meant to mention - I think this changes how the seed is used. I tried using a stable seed, but for some reason you get different results (though it's still stable). Does that make sense?

Basically, using `seed=x` on `master` gives different results to my branch, but repeatedly running my branch with `seed=x` will give the same result each time.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1193,8 +1193,33 @@ frequency = \case
...
+          []
+        (k, x) : xs ->
+          if n - k <= 0 then
+            []
+          else
+            x : upTo (n - k) xs
+
+    -- Pick a single generator from the weighted list of generators. We prune
+    -- here as shrinks no longer need to care about bias.
+    n <- prune $ integral $ Range.constant 1 total
```
HuwCampbell, line 1207: Instead of `integral` and `prune`, just use `integral_`.

ocharles: Thanks, I didn't know about that!

HuwCampbell:
Thanks Ollie. Cool investigation.

I've had a play and I think there's a slightly neater solution which doesn't require the tree hacking.

The intuition is the same, but just uses a modified shrink on integral to not include the duplicates.

```haskell
frequency :: MonadGen m => [(Int, m a)] -> m a
frequency = \case
  [] ->
    error "Hedgehog.Gen.frequency: used with empty list"
  xs0 -> do
    let
      pick n = \case
        [] ->
          error "Hedgehog.Gen.frequency/pick: used with empty list"
        (k, x) : xs ->
          if n <= k then
            x
          else
            pick (n - k) xs

      iis =
        scanl1 (+) (fmap fst xs0)

      total =
        last iis

    n <- shrink (const iis) $ integral_ $ Range.constant 1 total
    pick n xs0
```

What do you think?

ocharles:
That sounds great! I certainly wasn't happy about how many guts I had to open up to get this to work. I'll have a play around with your code tomorrow and see what I can make of it.

TysonMN:
> I think this changes how the seed is used. I tried using a stable seed, but for some reason you get different results (though it's still stable). Does that make sense?
> 
> Basically, using `seed=x` on `master` gives different results to my branch, but repeatedly running my branch with `seed=x` will give the same result each time.

Yes, that makes sense.  However, I don't understand why each case stopped where it did.  In your example, do all values correspond to a failed test?  That is, does the shrinking keeps happening until there is nothing left to shrink (not because the test passed)?

ocharles:
@TysonMN what you're seeing above is the output of `print` - https://hackage.haskell.org/package/hedgehog-1.0.3/docs/Hedgehog-Gen.html#v:print. This shows just one shrinking path. Perhaps you would like to see the output of `printTree` for an exhaustive enumeration of all shrinks? The way Hedgehog shrinking works is essentially a depth first traversal of the shrink tree, for as long as the test fails. The moment a test starts suceeding, we back track to the last failure and take the next alternative.

I forget exactly what `print` does, but I think it's essentially the left-most branch of the shrink tree.

ocharles:
@HuwCampbell Thanks for your suggestion, it looks like it works great!

```
> printTree $ frequency [(1, pure "a"), (3, pure "b"), (10, show <$> bool)]
"True"
 ├╼"a"
 ├╼"b"
 │  └╼"a"
 └╼"False"
```

Originally, it could be

```
"True"
 ├╼"a"
 ├╼"True"
 │  ├╼"a"
 │  ├╼"b"
 │  │  ├╼"a"
 │  │  └╼"b"
 │  │     └╼"a"
 │  ├╼"b"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  └╼"a"
 │  │  └╼"b"
 │  │     ├╼"a"
 │  │     └╼"b"
 │  │        └╼"a"
 │  └╼"False"
 ├╼"True"
 │  ├╼"a"
 │  ├╼"b"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  └╼"a"
 │  │  └╼"b"
 │  │     ├╼"a"
 │  │     └╼"b"
 │  │        └╼"a"
 │  ├╼"True"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  ├╼"a"
 │  │  │  └╼"b"
 │  │  │     └╼"a"
 │  │  ├╼"True"
 │  │  │  ├╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  ├╼"a"
 │  │  │  │  └╼"b"
 │  │  │  │     └╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  ├╼"a"
 │  │  │  │  ├╼"b"
 │  │  │  │  │  └╼"a"
 │  │  │  │  └╼"b"
 │  │  │  │     ├╼"a"
 │  │  │  │     └╼"b"
 │  │  │  │        └╼"a"
 │  │  │  └╼"False"
 │  │  └╼"False"
 │  └╼"False"
 ├╼"True"
 │  ├╼"a"
 │  ├╼"b"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  └╼"a"
 │  │  └╼"b"
 │  │     ├╼"a"
 │  │     └╼"b"
 │  │        └╼"a"
 │  ├╼"True"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  ├╼"a"
 │  │  │  └╼"b"
 │  │  │     └╼"a"
 │  │  ├╼"True"
 │  │  │  ├╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  ├╼"a"
 │  │  │  │  └╼"b"
 │  │  │  │     └╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  ├╼"a"
 │  │  │  │  ├╼"b"
 │  │  │  │  │  └╼"a"
 │  │  │  │  └╼"b"
 │  │  │  │     ├╼"a"
 │  │  │  │     └╼"b"
 │  │  │  │        └╼"a"
 │  │  │  └╼"False"
 │  │  └╼"False"
 │  ├╼"True"
 │  │  ├╼"a"
 │  │  ├╼"b"
 │  │  │  ├╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  └╼"a"
 │  │  │  └╼"b"
 │  │  │     ├╼"a"
 │  │  │     └╼"b"
 │  │  │        └╼"a"
 │  │  ├╼"True"
 │  │  │  ├╼"a"
 │  │  │  ├╼"b"
 │  │  │  │  ├╼"a"
 │  │  │  │  └╼"b"
 │  │  │  │     └╼"a"
 │  │  │  ├╼"True"
 │  │  │  │  ├╼"a"
 │  │  │  │  ├╼"b"
 │  │  │  │  │  ├╼"a"
 │  │  │  │  │  └╼"b"
 │  │  │  │  │     └╼"a"
 │  │  │  │  ├╼"b"
 │  │  │  │  │  ├╼"a"
 │  │  │  │  │  ├╼"b"
 │  │  │  │  │  │  └╼"a"
 │  │  │  │  │  └╼"b"
 │  │  │  │  │     ├╼"a"
 │  │  │  │  │     └╼"b"
 │  │  │  │  │        └╼"a"
 │  │  │  │  └╼"False"
 │  │  │  └╼"False"
 │  │  └╼"False"
 │  └╼"False"
 └╼"False"
```

so this is *much* better!

HuwCampbell:
Yeah I'm pretty happy with the result. It does address the duplicates seen in #118 as well.

HuwCampbell:
@ocharles can you squash this to one commit with both of us as authors?

ocharles:
@HuwCampbell How does that look?

HuwCampbell:
Thanks! that second shrink example is probably off now though.

ocharles:
I think it's still pretty reasonable, it might not be _exactly_ right, but it illustrates that there are no duplicates. I think the only material difference is that `Nothing` happens first, rather than last.

HuwCampbell:
I only mentioned because it was brought up that the initial value changed, while that's no longer the case.

ocharles:
Ah good point, I'll generate a new output.

ocharles:
Commit and pull request updated :writing_hand:

moodmosaic (approved):
Looks good 👍 Great collab work as well 🎸

PR #411, opened 2020-12-14
Generalise Hedgehog.Gen.element

ocharles:
Having this only work on lists is more restrictive than it needs to be. This commit generalizes `element` to instead work on any `Foldable`.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -1149,7 +1149,7 @@ constant =
 element :: (Foldable f, MonadGen m) => f a -> m a
 element fa = case toList fa of
   [] ->
-    error "Hedgehog.Gen.element: used with empty list"
+    error "Hedgehog.Gen.element: used with empty Foldable"
   xs -> do
     n <- integral $ Range.constant 0 (length xs - 1)
     pure $ xs !! n
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -1152,7 +1152,7 @@ element = \case
     error "Hedgehog.Gen.element: used with empty list"
   xs -> do
     n <- integral $ Range.constant 0 (length xs - 1)
-    pure $ xs !! toList n
+    pure $ toList xs !! n
 
 -- | Randomly selects one of the generators in the list.
 --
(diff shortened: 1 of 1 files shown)


ocharles:
I want this because in state machine testing I quite often do things like this:

```haskell
    commandGen :: State Symbolic -> Maybe (Gen (RegisterUser Symbolic))
    commandGen state = do
      knownUsernames <- nonEmpty $ state ^.. #users . folded . #username

      return do
        RegisterUser{ email, password } <- genRegisterUser
        username                        <- Gen.element $ toList knownUsernames

        return RegisterUser{ username, .. }
```

It's annoying to have to use `toList` here - it would be nice if I could just say:

```haskell
        username                        <- Gen.element knownUsernames
```

ocharles:
@jacobstanley this now actually builds :laughing:

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1146,8 +1146,8 @@ constant =
 --
 --   /The input list must be non-empty./
 --
-element :: MonadGen m => [a] -> m a
-element = \case
+element :: (Foldable f, MonadGen m) => f a -> m a
+element fa = case toList fa of
```
tmcgilchrist, line 1150: This could also use `null :: t a -> [Bool]` from Data.Foldable, the difference is probably negligible. 

The comment here should be "Hedgehog.Gen.element: used with empty foldable"

ocharles: Thanks, I've updated the error message. I'm not sure `null` would help much - we still need a list in order to use `!!` - so I've left that as is.

ocharles:
@moodmosaic any thoughts?

moodmosaic:
Thank you @ocharles 👍 🚀 

From the API consumer's point of view, `Gen.element $ knownUsernames` looks indeed betten than `Gen.element $ toList knownUsernames`.

A couple of notes:

- Types that are instances of Foldable include not only lists but also other data structures like trees, maps, sets, and more. Those aren't going to work in this case though...
- Perhaps accepting `NonEmpty a` would be best(?)
- [Postel's law](https://en.wikipedia.org/wiki/Robustness_principle) isn't violated by taking on Foldable, so that's a good thing

I've never seen a QC-style `elements` combinator accepting `Foldable`, that's why I added those notes above.

sol:
>> Perhaps accepting `NonEmpty a` would be best(?)

> It would definitely be an improvement.

I don't see this as a clear-cut improvement. `elements` is frequently used with static lists.

For static list you don't benefit much from the additional safety that `NonEmpty` provides, but it complicates your code.

(you would then need to use `NonEmpty.fromList` + add  the corresponding import)

moodmosaic:
Thank you, @ocharles. Also, thank you @tmcgilchrist and @sol for the comments.

PR #412, opened 2020-12-16
Replace HTraversable with TraversableB (from barbies)

### LINKED ISSUE #403: Replace HTraversable with Data.Functor.Barbie.TraversableB
While I've never been particularly excited about the names in the `barbies` library, it's still an excellent library. A benefit of using their `TraversableB` class (which is our `HTraversable`) is that it has a generic definition, so most things can be:

```haskell
data CommandInput f = CommandInput { foo, bar :: Var Int f }
  deriving (Generic, FunctorB, TraversableB)
```

and it *just works*. This is a pretty huge productivity boon!

ocharles:
Fixes #403


ocharles:
Open questions:

* Should we reexport `FunctorB` and `TraversableB`?

jacobstanley:
> * Should we reexport `FunctorB` and `TraversableB`?

I'm thinking yes, mostly because [HTraversable](https://hackage.haskell.org/package/hedgehog-1.0.4/docs/Hedgehog.html#t:HTraversable) is currently exposed.

hedgehog-example/hedgehog-example.cabal
```diff
@@ -53,6 +53,7 @@ library
       hedgehog
     , hedgehog-quickcheck
     , base                            >= 3          && < 5
+    , barbies
```
jacobstanley, line 56:
```suggestion
    , barbies                         >= 2.0        && < 2.1
```
(or whatever lower bound works)

mitchellwrosen:
An update - I believe 2.0 is more or less the correct `barbies` lower bound (as `Data.Functor.Barbie` did not exist prior), but (I think?) it would require dropping support for GHC 8.0 and 8.2 due to this:

```
src/Data/Generics/GenericN.hs:40:1: error:
    • Kind signature on data type declaration has non-* return kind k
    • In the data family declaration for ‘Param’
   |
40 | data family Param (n :: Nat) (a :: k) :: k
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

mitchellwrosen:
Ping!

jacobstanley:
> it would require dropping support for GHC 8.0 and 8.2 due to this

That's a shame but I think it's probably ok.

jacobstanley:
I found that the requires classes used to be in `Data.Barbie` and were moved to `Data.Functor.Barbie` in `barbie-2.0.0.0`

I've put in a module `Hedgehog.Internal.Barbie` which hides the required `CPP`.

PR #414, opened 2021-02-12
Build tree from values instead of wrapping and unwrapping

HuwCampbell:
This is just a minor cleanup, there should be no function differences.

As I mentioned earlier, my original version wrapped and unwrapped trees a bit much, making the code a bit harder to follow.

Pulling out a helper for `integral_` means we can build on values only.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -804,50 +804,50 @@ integral :: forall m a. (MonadGen m, Integral a) => Range a -> m a
 integral range =
   -- https://github.com/hedgehogqa/haskell-hedgehog/pull/413/files
   let
-    tryOriginFirst :: Tree.TreeT (MaybeT (GenBase m)) a -> Tree.TreeT (MaybeT (GenBase m)) a
-    tryOriginFirst tree =
-      Tree.TreeT $ do
-        let origin_ = Range.origin range
-        Tree.NodeT x xs <- Tree.runTreeT tree
-        pure $
-          if x == origin_ then
-            Tree.NodeT x xs
-          else
-            Tree.NodeT x (pure origin_ : xs)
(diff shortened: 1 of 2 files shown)


hedgehog/src/Hedgehog/Internal/Prelude.hs
```diff
@@ -8,10 +8,15 @@ module Hedgehog.Internal.Prelude (
...
 
 import           Control.Monad.Fail (MonadFail)
 
 import           Data.Semigroup (Semigroup(..))
 
 import           Prelude hiding (filter, print, map)
+
+(...) :: (c -> d) -> (a -> b -> c) -> a -> b -> d
+(...) = (.) . (.)
+{-# INLINE (...) #-}
```
jacobstanley, line 22: I think I'd rather not have this, this module was more intended for compatibility than introducing custom operators. I'd like to keep things as vanilla as possible for now.

Maybe you can convince me, but it shouldn't be part of this PR.

moodmosaic: @HuwCampbell, perhaps also we can move this to the let binding above where it’s used?

This clashes with ES6 spread syntax, so perhaps we could just rewrite it without (...) altogether.

moodmosaic (commented):
Great work! See @jacobstanley's comment and the one I also left. I'd be nice to have this PR on this release.

hedgehog/src/Hedgehog/Internal/Prelude.hs
HuwCampbell: Gone.  I've just eta expanded both arguments.

I do like the owl (called so as its implementation is `(.) . (.)`) as it's just another simple form of function composition, but I understand what you're going for with this module.

moodmosaic (approved):
👍

jacobstanley (approved):

PR #415, opened 2021-02-12
Don't shrink the action chosen in state machine testing

HuwCampbell:
While the contents of the action is fair game for shrinking, as is the sequence of actions, it makes little sense to change an action which may be the one which causes the error.


moodmosaic (approved):
👍

jacobstanley (approved):

PR #420, opened 2021-05-09
Add better TraversableB documentation including deprecation warning for HTraversable

jacobstanley:
Added an example showing how to derive `TraversableB` and a warning if you try to use `HTraversable`.

Hopefully that makes the upgrade process a bit easier.

Any suggestions welcome.

### DIFF
--- a/hedgehog-example/src/Test/Example/Registry.hs
+++ b/hedgehog-example/src/Test/Example/Registry.hs
@@ -23,7 +23,6 @@ import qualified Data.Set as Set
 
 import           Hedgehog
 import qualified Hedgehog.Gen as Gen
-import qualified Hedgehog.Internal.State as Gen
 import qualified Hedgehog.Range as Range
 
 import           System.IO.Unsafe (unsafePerformIO)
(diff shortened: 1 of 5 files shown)


ocharles:
Note that with `DeriveAnyClass` you don't need extra `instance` lines

jacobstanley:
It's true but `DeriveAnyClass` has other side effects which aren't ideal so I wasn't sure about recommending it.

Ideally we could do it with `DerivingVia`.

jacobstanley:
I originally had two examples, one using `DeriveAnyClass`, perhaps I'll go back to that.

ocharles:
Personally I advise my colleagues to use DeriveAnyClass with DervingStrategies. Best of all works then!

jacobstanley:
I think the syntax for `via` is a bit nicer, but I couldn't figure out how to write the instance for `Default` (or whatever we'd call it)

```hs
data Register v =
  Register Name (Var Pid v)
  deriving (Eq, Show, Generic)
  deriving (FunctorB, TraversableB) via (Default Register)
```

ocharles:
While I do agree, I think it might be best if a library like `hedgehog` avoids introducing something like `Default`. Let's just stick to what we're good at, which is property testing :)

jacobstanley:
Yeah I agree, I plan to add the `anyclass` example for this PR.

I don't think another solution is impossible longer term though, a few ideas:
- It could be another library, I use [quiet](https://hackage.haskell.org/package/quiet-0.2/docs/Quiet.html) for many of my show instances so that they're easier to read in tests.
- There is such a type [Default1](https://hackage.haskell.org/package/generic-deriving-1.14/docs/Generics-Deriving-Default.html#t:Default1) already but I think it has the wrong kind. If it was polykinded it would probably work. Of course you'd need the instance to be in `barbies` or `generic-deriving` then.

PR #421, opened 2021-05-13
Support shrinking with hedgehog generators in QuickCheck

HuwCampbell:
We can be more generous to folks with QuickCheck tests by allowing them to use hedgehog generators directly in their properties without sacrificing shrinking.

This is pretty explicity derived from @jacobstanley's code and QuickChecks's `forAll`.

I've run a bad example to show that it works fine with discards and predicates.
```
> let gInt = Gen.int (Range.constant 0 22)
> quickCheck $ Test.QuickCheck.Hedgehog.forAll (gInt >>= \i -> if even i then pure i else Gen.discard ) $ \x -> True Test.QuickCheck.=== (x < 18)
*** Failed! Falsified (after 2 tests and 2 shrinks):    
18
True /= False
```

### DIFF
--- a/hedgehog-quickcheck/src/Test/QuickCheck/Hedgehog.hs
+++ b/hedgehog-quickcheck/src/Test/QuickCheck/Hedgehog.hs
@@ -25,9 +25,9 @@ genSeed =
 
 -- | Create a QuickCheck 'QuickCheck.Gen' from a Hedgehog 'Gen'.
 --
---   /Note that this conversion does not preserve shrinking. To preserve
---    shrinking use the exported 'forAll' and 'forAllShow' functions rather
---    than going via a 'QuickCheck.Gen'./
+--   /Note that this conversion does not preserve shrinking. To preserve/
+--   /shrinking use the exported 'forAll' and 'forAllShow' functions rather/
+--   /than going via a 'QuickCheck.Gen'./
 --
 hedgehog :: Gen a -> QuickCheck.Gen a
 hedgehog gen =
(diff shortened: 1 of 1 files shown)


moodmosaic (commented):
Looks good :beers: Perhaps we need a different name than `forAll` so that it won't clash with QC.

HuwCampbell:
Yeah I considered that too, but thought one might import qualified (as we suggest with `Gen` and `Range`). I think it's fine either way.

jacobstanley:
I think we should rename it as the intended usage is as if the module belongs to QuickCheck and is part of its DSL.

(I agree it would make sense to follow `Gen` and `Result` practices if it was for use with the Hedgehog runner)

i.e.
```
-- unqualified
import Test.QuickCheck
import Test.QuickCheck.Hedgehog

-- or qualified the same as QuickCheck
import qualified Test.QuickCheck as QC
import qualified Test.QuickCheck.Hedgehog as QC
```

I can't think of any particularly satisfying names though, some ideas:

```
forAll genXXX $ \x ->
hedgehogAll genYYY $ \y ->

-- lol :boar: silly, but actually kinda good that it would line up with `forAll` when used
forAll genXXX $ \x ->
hogAll genYYY $ \y ->

-- I normally try to avoid single letter suffixes / use them as a last resort
-- but seeing this in context it's not so bad :thinking: 
-- (we do have forAllT in hedgehog proper also so it's not without precedent)
forAll genXXX $ \x ->
forAllH genYYY $ \y ->
```

HuwCampbell:
`forAllHedgehogs` 🦔

HuwCampbell:
I just went with the `H` suffix.

moodmosaic (approved):
I think the `H` suffix is reasonable in this context, what do you think @jacobstanley? :+1:

jacobstanley:
Yeah I think it's good 👍🏼

PR #422, opened 2021-05-18
Add `Num (GenT m a)` instance

i-am-tom:
Hey!

I've been writing a little simulation library, and a lot of the functions depend on delaying an action by some vague interval, and so I decided I could just sample `Gen` types. This has led to expressions like `exactly 2 days`, `roughly 6 hours`, and so on (where `exactly` is effectively the `pure` generator, and `roughly` does some magic with a normal distribution). Yesterday, I realised I wanted to simulate something that took roughly three days, but at least one day. So, I thought it would be neat to write `exactly 1 day + roughly 2 days`. This would require the typical `deriving via Ap (GenT m x)` instance for `GenT`, so here it is 😄

I hope I've followed the style guide, and I've tested this on 8.10.4, 8.6.3, and 8.4.3, though I assume Travis will tell me if I've missed anything! I couldn't find a set of per-instance tests to add to, but please do let me know if I've missed something.

Thanks so much for this library, it's invaluable!

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -507,7 +507,6 @@ instance (Monad m, Num a) => Num (GenT m a) where
     pure . fromInteger
 #endif
 
-
 instance Functor m => Functor (GenT m) where
   fmap f gen =
     GenT $ \seed size ->
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -235,6 +235,8 @@ import qualified Control.Monad.Fail as Fail
 #endif
 #if __GLASGOW_HASKELL__ < 806
 import           Data.Coerce (coerce)
+#else
+import           Data.Monoid (Ap (..))
 #endif
 
 ------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


moodmosaic (commented):
Thanks for the PR :+1: :heart: A few thoughts... Could this cause any weird/confusing type errors? Like if you made a typo, you type a number and ghc thinks you actually mean a generator.

i-am-tom:
Hm, good question! I _think_ as far as type errors go, this probably doesn't make anything too much worse? For example:

```haskell
Gen.alphaNum + 3
```

Before, this would complain about not finding an instance for `Num (Gen Char)`, but now it will complain about not finding an instance for `Num Char`. Perhaps the bigger worry is that something like this would now compile successfully:

```haskell
Gen.bool >>= \case
  True  -> Gen.list (Range.singleton 5) 3
  False -> pure []
```

Before, we'd get an error that boils down to not having an instance for `Num (Gen Int)` (or similar), whereas now we'll just take `3` here to mean `pure 3`. I'm not really sure where I stand on this: I feel like this probably less a typo and more an assumed behaviour, maybe? But I'm speaking totally from gut feeling at this point, so none of what I say should be trusted 😄

I'll leave it to you and your colleagues to make the call either way. If you can foresee some trouble for UX being introduced, I'm more than happy to newtype `Gen` for my own purposes and not get in the way!

jacobstanley:
>  If you can foresee some trouble for UX being introduced

Yeah I think this is a bit too much magic for me, thanks for the PR though ❤️

PR #423, opened 2021-06-23
Tweak 'cover' example code in haddocks

jhrcek:
This is a minor thing :smile:  but I was totally confused for few minutes after staring at this example in haddocks (notice how True and False are rendered - but the 2nd argument of Cover is `LabelName`)

Screenshot from 2021-06-23 09-01-43

This might be some haddock bug - it renders `String` literals in other code blocks correctly, but for some reason it renders True and False in this case as (dead) links.

After this small fix the example is rendered correctly (tested locally):

Screenshot from 2021-06-23 08-59-19


jacobstanley:
Thank you 🙏

Issue #432, opened 2021-09-08
Replace AppVeyor and Travis CI with GitHub actions


moodmosaic:
It'd be nice to replace AppVeyor (Windows builds) and Travis CI (Linux builds) with just a single cross-platform GitHub actions file.

@chshersh's [article](https://kodimensional.dev/github-actions) and @mrkkrp's [article](https://markkarpov.com/post/github-actions-for-haskell-ci.html) can be a good starting point for that.

moodmosaic:
Resolved via #434.

PR #434, opened 2021-09-19
Add Github Actions based CI

tmcgilchrist:
Includes Windows, linux (ubunutu) and OSX.Includes Windows, linux (ubunutu) and OSX. I think I've covered the same set of tests that travis was doing.

Notable differences to the Travis build is:
 * Only using cabal 3.4 (that could be added into the matrix at the cost of more builds)
 *  Using the `-latest` versions of each OS (https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)
     * windows-latest is Windows Server 2019
     * linux-latest is Ubuntu 20.04
     * macos-latest is Catalina 10.15

https://github.com/hedgehogqa/haskell-hedgehog/issues/432

Run results @ https://github.com/tmcgilchrist/haskell-hedgehog/pull/3/checks?check_run_id=3646820762


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
+name: CI
```
tmcgilchrist, line 1: You can call this workflow whatever you want here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,49 @@
...
+    branches:
+      - master
+  pull_request:
+    types:
+      - opened
+      - synchronize
+jobs:
+  build:
+    strategy:
+      fail-fast: false
```
tmcgilchrist, line 13: This will force all the builds to run, regardless of whether one of them fails. That isn't the default GH Action behaviour.

.github/workflows/ci.yaml
moodmosaic: I'll leave this up to @jacobstanley as he's better in naming things than I am 😆

.github/workflows/ci.yaml
moodmosaic: 👍

.github/workflows/ci.yaml
jacobstanley: CI is allg 😎

moodmosaic (approved):
👍

moodmosaic:
Thanks, @tmcgilchrist 🍻

tmcgilchrist:
No worries @moodmosaic

PR #439, opened 2021-11-27
Compatibility with the GHC 9.2 series

ysangkok:
This series ships with template-haskell version 2.18, which was previous
excluded by the bounds. The bounds have been bumped in this commit.

Also, this series has a new warning about defining superclasses in terms
of their subclasses. In this commit, we resolve the warning by moving
the definitions to the superclasses. Also, instead of duplicating a
`mappend` definition, we define it in terms of `(<>)` on its Semigroup
superclass.


ysangkok:
Oh, sorry, I hadn't thought there would be an existing PR for this. I'll close in favour of #436

PR #424, opened 2021-06-25
Apply patch for fixed seed with minor modifications.

simfleischman:
Mostly copied from
https://github.com/hedgehogqa/haskell-hedgehog/issues/201#issuecomment-441401915

PR #446, opened 2022-01-29
Apply patch for fixed seed with minor modifications. (updated)

jacobstanley:
I wasn't able to push to #424's branch so this is a rebase to check the build.

*edit: The 9.2 build found that a call to `read` resulted in a non-exhaustive pattern match so I have fixed that. Note that this means we now fail silently if `HEDGEHOG_SEED` can't be parsed. This is inline with how we handle the other environment variables which maybe isn't the best thing but is consistent at least. 😓*


jacobstanley:
Available in [hedgehog-1.1.1](https://hackage.haskell.org/package/hedgehog-1.1.1)

PR #453, opened 2022-02-28
Don't drop actions depending on shrunk predecessors.

### LINKED ISSUE #448: When a Command shrinks, later commands depending on it are dropped
...I think. Apologies if I'm misinterpreting what's going on, I'm not sure how to easily check.

I have a "setup" command, generating a Grid to insert into the database. Executing the action returns a GridId for it, which the update puts into the state along with the Grid. So the state is basically `Maybe (Grid, Var GridId v)`, and setup is only valid to generate when the state is `Nothing`.

Then I have a "validate" command, whose generator takes the GridId from the state, then exec reads some details about the Grid from the database and Ensure does some checks on it. Obviously this is only valid to generate when the state is `Just _`.

If I have a problem in the validation, I'd expect shrinking to reduce the Grid. But as far as I can tell, what actually happens is that it runs through every possible shrink of the Grid, inserting all of them in turn, passing every time, and finally fails with the sequence of actions (setup with non-minimal Grid, validate). My Grids have a lot of possible shrinks and my application code isn't well optimized, so this is slow.

My current guess is that when we shrink the setup action in the list, after shrinking the user-provided generator, `contextNewVar` creates a new variable to put in the context instead of reusing the existing one. If that's right, then `dropInvalid` would see that the validate action is referring to a `Var` that doesn't exist any more, and remove it. So we might overall try the action sequences

* setup 100 -> var 1, validate (var 1) -> var 2
    * This fails.
* setup 99 -> var 3, ~~validate (var 1) -> var 2~~
    * The validate refers to a var that doesn't exist. So that gets dropped, and we just run the setup, which passes.
* setup 98 -> var (3 or 4?), ~~validate (var 1) -> var 2~~
    * Validate gets dropped for the same reason, so pass again.
* and so on, until the smallest possible `setup` passes, and we go back to the first one we tried.

More speculatively, if I'm right about all this, I wonder if we could solve the problem by rearranging [the definition of `action`](https://github.com/hedgehogqa/haskell-hedgehog/blob/f3473dfb4ea817222a09033a90d15bd30ce07f89/hedgehog/src/Hedgehog/Internal/State.hs#L544) slightly. `contextNewVar` gets called after we use the generator. I wonder whether moving the line `output <- contextNewVar` above `input <- ...` would be sufficient. Then (according to my very limited understanding) we'd reuse the Var after shrinking instead of creating a new one. A downside would be that if `Require` fails, we've still put the Var in the context; I don't know if that's a significant problem. I also don't know if there'd be other problems with it.

If this doesn't seem right and also isn't enough detail to figure out what's going on instead, I can investigate more deeply and try to get a minimal example.

ChickenProp:
Closes #448. Suppose we have an Action list

    A 1 -> Var 0
    B (Var 0) -> ...

Then when we shrink A, we would previously get the list

    A 0 -> Var 1
    B (Var 0) -> ...

And then we'd drop B from this list, because `Var 0` no longer exists.
Now shrinking will give

    A 0 -> Var 0
    B (Var 0) -> ...

which is fine.

This means we now generate Vars even for actions whose `Require` fails.

I don't know if this is the ideal way to fix this. I wonder whether, when shrinking, the context should be rolled back so that the output we previously generated is no longer present? But I wouldn't know how to change that, and I'd be much less confident of not breaking anything else if I did.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/State.hs
+++ b/hedgehog/src/Hedgehog/Internal/State.hs
@@ -548,6 +548,12 @@ action commands =
     Command mgenInput exec callbacks <-
       Gen.element_ $ filter (\c -> commandGenOK c state0) commands
 
+    -- If we shrink the input, we still want to use the same output. Otherwise
+    -- any actions using this output as part of their input will be dropped. But
+    -- the existing output is still in the context, so `contextNewVar` will
+    -- create a new one. To avoid that, we generate the output before the input.
+    output <- contextNewVar
+
     input <-
       case mgenInput state0 of
         Nothing ->
@@ -559,8 +565,6 @@ action commands =
       pure Nothing
 
     else do
-      output <- contextNewVar
-
       contextUpdate $
         callbackUpdate callbacks state0 input (Var output)
 


jacobstanley:
This is a good catch.

The vars end up being synonymous with the index of the action in the final list, it doesn't really make sense to change that during shrinking.

jacobstanley:
> This means we now generate Vars even for actions whose `Require` fails.

It might make sense in the long run to have a separate pass which minimizes the variable ids. That is, if you're left with three variables after shrinking they are always `0`, `1`, `2`.

moodmosaic:

PR #455, opened 2022-04-13
Clarify that subsequences are strictly smaller

sshine:
Considering `Data.List.subsequences` contains its own input:

```
> Data.List.subsequences [1,2,3]
[[],[1],[2],[1,2],[3],[1,3],[2,3],[1,2,3]]
```

it can be clarified that `Hedgehog.Gen.subsequence` *also* does ~not~.


ChickenProp:
Does it not? It looks like it includes each element with probability 1/2, so for a list of length n, it should generate the whole list with probability 1/2^n. Am I missing something?

sshine:
You're right, I am mistaken. Thanks for noticing.

Since `Data.List.subsequences` does not explicate, perhaps neither should this one.

PR #449, opened 2022-02-20
Compatibility with text-2.0

ysangkok:
Tested using

    cabal test --constraint='text>=2' all -w ghc-9.2.1

Fixes #443.

The test failures of old GHC versions on Windows are compiler bugs, it seems: haskell/text/pull/404

PR #427, opened 2021-07-17
Allow mmorph 1.2

felixonmars:
Builds fine and all tests pass.

PR #457, opened 2022-05-21
Allow newer dependencies

ysangkok:
Tested with the following addition to `cabal.project` (to fix `happy`):

```
source-repository-package
  type: git
  location: git@github.com:haskell/happy.git
  tag: cc0fc789a19b9c3bdb1b1ed29bf910f386cefdf7
  subdir: packages/tabular

source-repository-package
  type: git
  location: git@github.com:haskell/happy.git
  tag: cc0fc789a19b9c3bdb1b1ed29bf910f386cefdf7
  subdir: packages/grammar
```

And the following command line:

    cabal test all --constraint='mtl>=2.3' --constraint='transformers>=0.6' --constraint='mmorph>=1.2'

All test pass.

I don't think the happy override should be committed since it will be unnecessary soon when haskell/happy#236 is closed.

This follows up on bumps missed in #427 and #449.

### DIFF
--- a/hedgehog-example/hedgehog-example.cabal
+++ b/hedgehog-example/hedgehog-example.cabal
@@ -59,8 +59,8 @@ library
     , filepath                        >= 1.3        && < 1.5
     , hashtables                      >= 1.2        && < 1.4
     , lifted-base                     >= 0.2        && < 0.3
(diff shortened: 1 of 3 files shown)


jacobstanley:
Thanks! 🎉

PR #461, opened 2022-08-14
GHC 9.4 compatibility

ysangkok:
This allows compatibility with GHC 9.4. Tested with `cabal test -w ghc-9.4.1`.

The extension is added because of this new warning:

```
src/Hedgehog/Internal/Gen.hs:312:46: error: [-Wtype-equality-requires-operators, -Werror=type-equality-requires-operators]
    The use of ‘~’ without TypeOperators
    will become an error in a future GHC release.
    Suggested fix: Perhaps you intended to use TypeOperators
    |
312 | toTree :: forall m a. (MonadGen m, GenBase m ~ Identity) => m a -> m (Tree a)
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,8 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         os:    [macos-latest, ubuntu-latest, windows-latest]
-        cabal: ["3.4"]
-        ghc:   ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.1"]
+        cabal: ["3.8.1.0"]
+        ghc:   ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.4", "9.4.1"]
         exclude:
           # https://github.com/haskell/text/pull/404
           - os: windows-latest
--- a/hedgehog-example/hedgehog-example.cabal
+++ b/hedgehog-example/hedgehog-example.cabal
@@ -66,7 +66,7 @@ library
     , process                         >= 1.2        && < 1.7
     , QuickCheck                      >= 2.7        && < 2.15
     , resourcet                       >= 1.1        && < 1.3
-    , template-haskell                >= 2.10       && < 2.19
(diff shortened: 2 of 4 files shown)


moodmosaic:
Thank you, @ysangkok :+1: Shouldn't we also add an entry in the CI build for 9.4.1?

moodmosaic (commented):
Thank you, @ysangkok :+1: Shouldn't we also add an entry in the CI build for 9.4.1?

moodmosaic:
Hmm... It looks like CI failed to fetch Cabal 3.8 (required by GHC 9.4). That could be an issue with GHCUp or with the (no longer maintained) [actions/setup-haskell](https://github.com/actions/setup-haskell).

moodmosaic:
@ysangkok let's see if the force-pushed changes you pushed will go through (I've noticed you removed Cabal 3.8, but IIRC that's the minimum required version that'll work with GHC 9.4). :eyes:

ysangkok:
@moodmosaic I *replaced* Cabal 3.4 with Cabal 3.8. The failure was because GHCup apparently requires the full version number, I don't think it was related to setup-haskell. The build passes now, I think this is good to go.

.github/workflows/ci.yaml
```diff
@@ -13,8 +13,8 @@ jobs:
       fail-fast: false
       matrix:
         os:    [macos-latest, ubuntu-latest, windows-latest]
-        cabal: ["3.4"]
-        ghc:   ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.1"]
+        cabal: ["3.8.1.0"]
+        ghc:   ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.4", "9.4.1"]
```
moodmosaic, line 17: 👍👌

moodmosaic (approved):
🦔🐾👍

moodmosaic:
> @moodmosaic I _replaced_ Cabal 3.4 with Cabal 3.8.

Yep! :+1: That's [fine](https://www.haskell.org/ghc/download_ghc_9_4_1.html).

moodmosaic:

PR #466, opened 2022-09-10
Replace `exceptions` dependency with `safe-exceptions`

ocharles:
By using `safe-exceptions` we avoid catching asynchronous exceptions, which aren't of any interest to Hedgehog. The current usage of `exceptions` is problematic, because `checkReport` calls `catchAny`, which means it catches asynchronous exceptions. This is a problem for `tasty-hedgehog`, because it means that Tasty's timeout mechanism can't be used to timeout Hedgehog tests (https://github.com/qfpl/tasty-hedgehog/issues/64, https://github.com/UnkindPartition/tasty/issues/349)


hedgehog/src/Hedgehog/Internal/Exception.hs
```diff
@@ -5,7 +5,7 @@ module Hedgehog.Internal.Exception (
   ) where
 
 import           Control.Exception (Exception(..), AsyncException, SomeException(..), evaluate)
-import           Control.Monad.Catch (MonadCatch(..), throwM)
+import           Control.Exception.Safe (MonadCatch(..), catch, throwM)
```
moodmosaic, line 8:
```
Error:     The import item ‘MonadCatch(..)’ suggests that
    ‘MonadCatch’ has (in-scope) constructors or class methods,
    but it has none
  |
8 | import           Control.Exception.Safe (MonadCatch(..), catch, throwM)
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

ocharles: Yea, I've seen the CI failures, but I wanted to get feedback as to whether this would be accepted before continuing.

moodmosaic (commented):
Thank you @ocharles 👍

This PR looks reasonable, given the nature of async exceptions (coming outside of the current execution) which is something we never want to recover. Good to know that this helps `tasty`.

I just left a comment in the PR.

ocharles:
Great! I'll get these CI problem fixed up and then I'll squash it down to one commit.

ocharles:
@moodmosaic Ok, all done! I think this is ready to go now.

moodmosaic:
Thank you, @ocharles 👍

ocharles:
YW!

PR #468, opened 2022-09-15
Document sample

### LINKED ISSUE #467: Document `sample` with caveats and warnings
Our codebase test suite is littered with `Gen.sample`, which has a bunch of negative properties:

1. `Gen.sample` doesn't accept a seed and can't be made deterministic from a CLI argument
2. People think they're getting a property test, when they're really getting a non-deterministic random unit test
3. It's slower than just having a bit of data around as a fixture

I can't really blame them - we recommend `hoogle` and if you search `Gen a -> IO a`, `sample` pops up first. The documentation for `sample` mentions none of these issues:

Screenshot from 2022-09-15 09-00-45

IMO this function needs to be documented like "This is only for use with GHCi and other quick examples. Please do not use it directly in your test suite - use the `forAll` method to turn a `Gen` into a `PropertyT` and test with that."

parsonsmatt:
This PR adds documentation to the `sample` function and provides a warning that it is not suitable for testing with.

FIxes #467

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -1744,6 +1744,10 @@ shuffleSeq xs =
 
 -- | Generate a sample from a generator.
 --
+-- This function is useful for examining a 'Gen' in GHCi or other contexts.
+-- It is not appropriate for use in a test suite directly. You will only
+-- get a single sample from this function, and it will not give you
+-- a property test. The seed is random, so the test is not deterministic.
 sample :: MonadIO m => Gen a -> m a
 sample gen =
   liftIO $


hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1744,6 +1744,10 @@ shuffleSeq xs =
 
 -- | Generate a sample from a generator.
 --
+-- This function is useful for examining a 'Gen' in GHCi or other contexts.
+-- It is not appropriate for use in a test suite directly. You will only
+-- get a single sample from this function, and it will not give you
+-- a property test. The seed is random, so the test is not deterministic.
```
moodmosaic, line 1750: 👍

While we're here, perhaps it'd be nice to expand and provide some alternative(s) also?

moodmosaic (commented):
Thanks, @parsonsmatt! Left a comment there.

hedgehog/src/Hedgehog/Internal/Gen.hs
```diff
@@ -1748,6 +1748,16 @@ shuffleSeq xs =
 -- It is not appropriate for use in a test suite directly. You will only
 -- get a single sample from this function, and it will not give you
 -- a property test. The seed is random, so the test is not deterministic.
+--
+-- If you only want a single test to run, then user @'withTests' 1@:
```
ocharles, line 1752:
```suggestion
-- If you only want a single test to run, then use @'withTests' 1@:
```

moodmosaic (approved):
👍

Issue #472, opened 2022-12-22
Should size grow faster when the number of tests is small?


ChickenProp:
I often don't want to run a full set of tests and use `NoConfidenceTermination 10`. (Specifically I use `modifyMaxSuccess (const 10)` from hspec-hedgehog). But that means the size never grows above 9, so `Range.linear 0 10` will always generate 0, which isn't what I expected.  It works out okay if I write my generators expecting a specific number of max-tests, or I can call `Gen.scale` everywhere. But it would be more convenient if the size would grow to something large in a small number of tests.

Quickcheck does [something like this](https://github.com/nick8325/quickcheck/blob/71668ed61ef6f5fdb7ed8cff8a14e3e4233829b3/src/Test/QuickCheck/Test.hs#L226). I don't fully understand the arithmetic but in this case the size would go 0, 10, 20, ..., 90. (`n` is the test number and `d` is a discard count.)

I don't think it has an analogue of `EarlyTermination` or `NoEarlyTermination`, so hedgehog would need to figure out what to do about that. I don't use these myself, but from what I gather from the code:

* `NoEarlyTermination c t` still runs exactly `t` tests, the same as `NoConfidenceTermination t`, so it could have the same behavior as `NoConfidenceTermination`.
* `EarlyTermination c t` actually ignores `t` and runs at least 100 tests, more if needed for the confidence threshold `c`, but always some multiple of 100. (Is this intended? It's not what the name suggests to me.) So it could just keep the existing behavior.

So these don't seem like big problems to me.

PR #474, opened 2023-01-30
Grow size faster with a small number of tests

ChickenProp:
As described in https://github.com/hedgehogqa/haskell-hedgehog/issues/472. We now grow test size in fixed increments from 0 to 99, or as close as we can without going over, in however many tests we run. If we run over *n* * 100 tests, then we go from 0 to 99 *n* times, and then do fixed increments for the remainder. Additionally, if we discard a bunch of times in a row we start to grow the size.

This will be helpful for https://github.com/proda-ai/excelsior/pull/8811, where currently we do 10 tests at a time and use sizes that would be larger than intended if we did more.

I've removed `recheck` because "specify the starting size" doesn't really make sense with this, and that function is no longer suggested by failure reports. If this goes upstream I dunno if they'll want to keep it for backwards compatibility, but we can figure something out if so.

### DIFF
--- a/hedgehog/hedgehog.cabal
+++ b/hedgehog/hedgehog.cabal
@@ -137,6 +137,7 @@ test-suite test
     Test.Hedgehog.Filter
     Test.Hedgehog.Maybe
     Test.Hedgehog.Seed
+    Test.Hedgehog.Size
     Test.Hedgehog.Skip
     Test.Hedgehog.Text
     Test.Hedgehog.Zip
--- a/hedgehog/src/Hedgehog.hs
+++ b/hedgehog/src/Hedgehog.hs
@@ -59,7 +59,6 @@ module Hedgehog (
   , discard
 
   , check
-  , recheck
   , recheckAt
 
   , discover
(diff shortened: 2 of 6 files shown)


ChickenProp:
So the exact implementation here is incompatible with #489, because it means the exact sequence of tests and discards is now relevant. Currently it affects neither the seed nor the size that we run the next test at, but with this PR it can affect the size. If we've skipped 10 failing tests and 10 discards, we'll be running one size if the sequence was (T×10, D×10) and a different size if it was anything else.

A few possible approaches:

1. Save the exact sequence of tests and discards in the skip. This would make the compressed string longer and harder to read, but maybe that's fine. Might be the most robust to future changes, allowing the exact sequence to affect both seed and size.
2. Save the final size in the skip (maybe `testCount[/discardCount]@size[:path]`). Still relies on the exact sequence not changing seed, but allows it to change size.
3. Save the number of recent discards in the skip (maybe `testCount[/discardCount[+recentDiscardCount]][:path]`). Makes assumptions about what features of the sequence can change the size.
4. Pass `calculateSize` the number of total discards, and stop tracking the number of recent discards. Instead of loosening the "exact sequence doesn't affect size" assumption, we simply make that assumption true again. This would make size monotonic (except for resetting to 0 every 100 tests), which the current implementation isn't.
 
All of these seem basically okay, except I think (3) doesn't really have anything to recommend it over (2). If this feature is wanted, I can implement any of them; my recommendation would be (2) or maybe (4). If it's not, then I'll probably implement (4) on my fork next time I merge master, just because it's easiest.

PR #475, opened 2023-01-30
Grow size faster with a small number of tests

ChickenProp:
As described in https://github.com/hedgehogqa/haskell-hedgehog/issues/472. We now grow test size in fixed increments from 0 to 99, or as close as we can without going over, in however many tests we run. If we run more than *n* * 100 tests, then we go from 0 to 99 *n* times, and then do fixed increments for the remainder. Additionally, if we discard a bunch of times in a row we start to grow the size.

This will be helpful for https://github.com/proda-ai/excelsior/pull/8811, where currently we do 10 tests at a time and use ranges that would be larger than intended if we did more. We'll want to update stack.yaml to use this, and narrow the ranges we specify.

I've removed `recheck` because "specify the starting size" doesn't really make sense with this, and that function is no longer suggested by failure reports. If upstream wants to keep it for backwards compatibility we can figure something out.

Upstream PR is https://github.com/hedgehogqa/haskell-hedgehog/pull/474.


ChickenProp:
Sorry, this definitely shouldn't have been targeted at this repo.

PR #481, opened 2023-03-14
Support GHC-9.6

### LINKED ISSUE #480: Support GHC-9.6 (template-haskell-2.20)
The current revision of hedgehog-1.2 rejects template-haskell-2.20 which is the version shipped with GHC-9.6. It would be nice if hedgehog supported this version of template-haskell such that this new GHC version is supported.

ysangkok:
Fixes #480

### DIFF
--- a/hedgehog-example/hedgehog-example.cabal
+++ b/hedgehog-example/hedgehog-example.cabal
@@ -66,7 +66,7 @@ library
     , process                         >= 1.2        && < 1.7
     , QuickCheck                      >= 2.7        && < 2.15
     , resourcet                       >= 1.1        && < 1.4
-    , template-haskell                >= 2.10       && < 2.20
+    , template-haskell                >= 2.10       && < 2.21
     , temporary                       >= 1.3        && < 1.4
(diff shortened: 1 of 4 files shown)


moodmosaic:
Thank you! 🎯👍

ysangkok:
@moodmosaic Should I make a PR to remove the 8.0 and 8.2 builds, or should they get fixed?

ysangkok:
@moodmosaic I found out why the build was failing! The reason is that newer versions of `text` and `vector` don't work with the Haddock documentation generating shipped with older versions of GHC. So we don't need to remove these GHC versions, we just need to disable Haddocks for them. I reported submitted #482 which fixes the CI. That PR also links to the upstream bug report that I submitted. They are unwilling to work around this issue upstream, so we have to do it here.

moodmosaic:
Awesome, @ysangkok. That seems to be the best option, after all:
>we don't need to remove these GHC versions entirely, we just need to disable Haddocks for them

Thank you 👍

PR #482, opened 2023-03-24
Update Github CI actions, exclude Haddocks for old GHCs

ysangkok:
GHC 8.0, 8.2 and 8.4 failed because these old GHC versions ship a buggy Haddock. I added a predicate to skip Haddock generation on these old versions. Also bumped all other bounds old versions I could find.

See 
* https://github.com/haskell/text/issues/514


moodmosaic:
Thank you, @ysangkok, for investigating 👍 💯

PR #488, opened 2023-05-23
Add shrink timeout

### LINKED ISSUE #476: Shrink with a timeout
The current way to control shrinking is with a maximum number of shrinks. It's really hard to tweak this, because I don't really have any intuition behind that number - only "more leads to a better shrink". What I'd really rather do is shrink for up to 30s, and after that just show me the best counterexample that could be found.

tbidne:
Resolves https://github.com/hedgehogqa/haskell-hedgehog/issues/476.

There are some design decisions to make, so this is intended more to jumpstart the conversation than it is to present a finished implementation. Some notes:

1. I chose to add a new field, `ShrinkTimeLimit :: Int` independent of the current `ShrinkLimit :: Int`. This way is backwards compatible, and it allows one to specify both a total number of shrinks and a time limit. That said, it is arguable that combining these choices into a single option makes for a friendlier interface as e.g. someone might specify a time limit of `30s` only to be unexpectedly thwarted by the default `ShrinkLimit` of 1,000. 
1. Bikeshedding: `ShrinkTimeLimit` could also be named `ShrinkTimeout`.
1. The field is `Int` microseconds to make interop with [`timeout`](https://hackage.haskell.org/package/base-4.18.0.0/docs/System-Timeout.html#v:timeout) simpler, though I could imagine choosing something less implementation-derived, say, seconds (and maybe a different type e.g. `Natural`).
1. In case of a timeout, we want to present the latest shrink, so I modified the shrink loop to take in an "update" parameter that saves the current result in an `IORef` iff `ShrinkTimeLimit` is set. If `ShrinkTimeLimit` is not set then the "update" logic is `const (pure ())`. I mention this in case it can affect performance, as an alternative would be to have two totally different loops, at the cost of code reuse.
1. I added some tests, though the timeout test is naturally pretty weak as it is non-deterministic. If anyone has ideas here, I'm all ears.

Thanks!

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Runner.hs
+++ b/hedgehog/src/Hedgehog/Internal/Runner.hs
@@ -123,39 +123,35 @@ takeSmallest ::
   -> ShrinkPath
   -> ShrinkLimit
   -> ShrinkRetries
-  -> (Progress -> m ())
-  -> NodeT m (Maybe (Either Failure (), Journal))
-  -> m Result
+  -> (FailureReport -> m ())
+  -> Failure
+  -> Journal
+  -> [TreeT m (Maybe (Either Failure (), Journal))]
+  -> m FailureReport
 takeSmallest shrinks0 (ShrinkPath shrinkPath0) slimit retries updateUI =
   let
-    loop shrinks revShrinkPath = \case
-      NodeT Nothing _ ->
-        pure GaveUp
+    loop shrinks revShrinkPath (Failure loc err mdiff) (Journal logs) xs = do
+      let
+        shrinkPath =
+          ShrinkPath $ reverse revShrinkPath
+        failure =
+          mkFailure shrinks shrinkPath Nothing loc err mdiff (reverse logs)
 
-      NodeT (Just (x, (Journal logs))) xs ->
-        case x of
-          Left (Failure loc err mdiff) -> do
-            let
-              shrinkPath =
-                ShrinkPath $ reverse revShrinkPath
-              failure =
-                mkFailure shrinks shrinkPath Nothing loc err mdiff (reverse logs)
+      updateUI failure
 
-            updateUI $ Shrinking failure
+      if shrinks >= fromIntegral slimit then
+        -- if we've hit the shrink limit, don't shrink any further
+        pure failure
+      else
+        findM (zip [0..] xs) failure $ \(n, m) -> do
+          o <- runTreeN retries m
+          case o of
+            NodeT (Just (Left smallerFailure, smallerLogs)) children ->
+              Just <$>
+                loop (shrinks + 1) (n : revShrinkPath) smallerFailure smallerLogs children
+            _ ->
+              return Nothing
 
-            if shrinks >= fromIntegral slimit then
-              -- if we've hit the shrink limit, don't shrink any further
-              pure $ Failed failure
-            else
-              findM (zip [0..] xs) (Failed failure) $ \(n, m) -> do
-                o <- runTreeN retries m
-                if isFailure o then
-                  Just <$> loop (shrinks + 1) (n : revShrinkPath) o
-                else
-                  return Nothing
-
-          Right () ->
-            return OK
   in
     loop shrinks0 (reverse shrinkPath0)
 
@@ -348,25 +344,27 @@ checkReport cfg size0 seed0 test0 updateUI = do
                   Report (tests + 1) discards coverage0 seed0
               mkReport <$> skipToShrink shrinkPath (updateUI . mkReport) node
             _ -> do
-              node@(NodeT x _) <-
+              NodeT x trees <-
                 runTreeT . evalGenT size s0 . runTestT $ unPropertyT test
               case x of
                 Nothing ->
                   loop tests (discards + 1) (size + 1) s1 coverage0
 
-                Just (Left _, _) ->
+                Just (Left failure, journal) ->
                   let
                     mkReport =
                       Report (tests + 1) discards coverage0 seed0
                   in
-                    fmap mkReport $
+                    fmap (mkReport . Failed) $
                       takeSmallest
                         0
                         (ShrinkPath [])
                         (propertyShrinkLimit cfg)
                         (propertyShrinkRetries cfg)
-                        (updateUI . mkReport)
-                        node
+                        (updateUI . mkReport . Shrinking)
+                        failure
+                        journal
+                        trees
 
                 Just (Right (), journal) ->
                   let


ChickenProp:
Not a maintainer myself, but I think this is neat!

More name bikeshedding: I'd suggest adding "micros" (e.g. `withShrinkTimeMicros`, `ShrinkTimeLimitMicros`) to make it clear what the units are.

> I added some tests, though the timeout test is naturally pretty weak as it is non-deterministic. If anyone has ideas here, I'm all ears.

Given a known shrink tree, you could have a test deliberately hang on a given input, e.g. a timeout of 100 microseconds and hang for 200 after five shrinks.

TysonMN:
My two cents:

I like the naming format `xInUnits`.  In this case, the name would be `withShrinkTimeInMicroseconds`.  I admit that the name can get a bit long though.

tbidne:
> @ChickenProp Given a known shrink tree, you could have a test deliberately hang on a given input, e.g. a timeout of 100 microseconds and hang for 200 after five shrinks.

Thanks for the idea! I added such a test [here](https://github.com/hedgehogqa/haskell-hedgehog/pull/488/files#diff-e85365fad77027a2c0b3feb71d6973d7e6e5b0270bc94fd1e2a99cdff1b41ebaR65). I made the timeout longer (1 second) as lower values would often cause at least one ci job to fail (usually MacOS on an older GHC), due to the test timing out before it reached the generated values I specifically wanted it get stuck on.

I also had a "sanity-check" style test that verified the timeout indeed cancels shrinking in the expected wall-clock time.

```haskell
-- Time limit of 2 seconds. Verifies that withShrinkTime indeed cancels
-- shrinking within the time limit we want.
prop_ShrinkTimeLimitClock :: Property
prop_ShrinkTimeLimitClock =
  property $ do
    startTime <- liftIO $ Clock.getMonotonicTime
    annotateShow startTime
    _ <- checkModPropGen delay30s (withShrinkTime 2000000)
    endTime <- liftIO $ Clock.getMonotonicTime
    annotateShow endTime
    let timeElapsed = endTime - startTime
    annotateShow timeElapsed
    -- should be around 2
    diff timeElapsed (>=) 1.5
    diff timeElapsed (<=) 2.5
  where
    delay30s x = when (x == 13) (liftIO $ CC.threadDelay 30000000)
```

Unfortunately this relies on `GHC.Clock`, which requires a newer GHC (8.4.1) than the oldest on CI (8.0.2). I suppose I could conditionally add it with cpp, or maybe this test isn't so important anyway.

@TysonMN

As far as the name goes, I agree adding the units is a good idea. Yet `ShrinkTimeInMicroseconds` is a bit grim, considering the corresponding functions/fields e.g. `propertyShrinkTimeInMicroseconds`. Perhaps `ShrinkTimeoutMicros`? I don't care too much about `InMicro...` vs. `Micro...`, but I think `Microseconds` should probably be abbreviated (`Micro` or `Micros`), and `ShrinkTimeout` sounds a little better to me than `ShrinkTimeLimit`.

tbidne:
Also it is probably a good idea to mention `withShrinks` in the documentation for `withShrinkTime` e.g.

```haskell
-- | Set the timeout -- in microseconds -- after which the test runner gives up
-- on shrinking and prints the best counterexample. Note that shrinking can be
-- cancelled before the timeout if the 'ShrinkLimit' is reached
-- (defaults to 1,000). See 'withShrinks'.
--
withShrinkTime :: ShrinkTimeLimit -> Property -> Property
```

tbidne:
I went ahead and made those changes:

- Renamed `ShrinkTimeLimit --> ShrinkTimeoutMicros`.
- Added wall clock test w/ cpp.

moodmosaic:
@tbidne, can I re-review this?

tbidne:
@moodmosaic Please do, thanks!

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -118,46 +123,75 @@ runTreeN n m = do
...
+    Nothing -> runLoop (const (pure ()))
+    -- run the loop in the timeout
+    Just (ShrinkTimeoutMicros timeoutMicros) -> \nodeT -> do
+      resultSoFar <- liftIO $ newIORef Nothing
+      let updateResultSoFar = liftIO . writeIORef resultSoFar . Just
+      timeout timeoutMicros (runLoop updateResultSoFar nodeT) >>= \case
+        -- timed out, return preliminary result if it exists
+        Nothing -> liftIO (readIORef resultSoFar) <&> \case
+          Nothing -> GaveUp
+          Just r -> r
```
HuwCampbell:
```haskell
fromMaybe GaveUp <$> liftIO (readIORef resultSoFar)
```
Then the extra operator isn't required.

HuwCampbell: As you're already in a `do` block, I would also suggest just `<-` binding the result of timeout then doing a normal case statement on the next line. But that's pretty minor.

HuwCampbell:
I'm not a maintainer on the Haskell package, but it looks ok to me apart from a few small things.

The [style guide](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md?plain=1#L145) is pretty clear on its preference for `let` over `where` (and I left a few other comments).

I'm not sure the behaviour with small timeouts is correct. It should report the failure case which started the search instead of `GaveUp` if a smaller failing case can't be found within the time limit.

Timing based tests are always a bit finicky – maybe add something to the examples instead?

----

Having thought about this a bit more; I think that the current type for `takeSmallest` and `loop` is actually bad, and if that gets fixed first it'll improve things markedly.

The "tell" is `isFailure` and `Just (Left _, _)`. We're throwing away information (validating, not parsing).

There's also weirdness around its returns. It can't return `GaveUp` or `Ok` in practice; but that's not reflected in its type.

Thus: I think the shrink should be:
```haskell
takeSmallest ::
     MonadIO m
  => ShrinkCount
  -> ShrinkPath
  -> ShrinkLimit
  -> ShrinkRetries
  -> (FailureReport -> m ())
  -> Failure
  -> Journal
  -> [TreeT m (Maybe (Either Failure (), Journal))]
  -> m FailureReport
takeSmallest shrinks0 (ShrinkPath shrinkPath0) slimit retries updateUI =
  let
    loop shrinks revShrinkPath (Failure loc err mdiff) (Journal logs) xs = do
      let
        shrinkPath =
          ShrinkPath $ reverse revShrinkPath
        failure =
          mkFailure shrinks shrinkPath Nothing loc err mdiff (reverse logs)

      updateUI failure

      if shrinks >= fromIntegral slimit then
        -- if we've hit the shrink limit, don't shrink any further
        pure failure
      else
        findM (zip [0..] xs) (failure) $ \(n, m) -> do
          o <- runTreeN retries m
          case o of
            NodeT (Just (Left smallerFailure, smallerLogs)) children ->
              Just <$>
                loop (shrinks + 1) (n : revShrinkPath) smallerFailure smallerLogs children
            _ ->
              return Nothing

  in
    loop shrinks0 (reverse shrinkPath0)
```

Ok. Back to the PR at hand, if you use that type, then things get a bit simpler, as you have "on hand" the starting position and won't return `GaveUp`.

There's also the question of multiple return paths, the `IORef`, and the returned value. If they're always in sync, why have both?

moodmosaic:
Thanks @HuwCampbell 👍 Agreed! 

@tbidne could you address the above? It LGTM once the comments are addressed. Thank you! 🙏

tbidne:
Apologies for the delay.

I have rebased the PR via @HuwCampbell's excellent suggestion. The first commit is purely the suggested refactor, which passes the test suite locally.

The second commit is my changes, and indeed the refactor makes this significantly simpler. I also added the other requested changes e.g. `let vs. where` and moving the tests to `examples`. It may therefore be easier to review the commits individually.

I am curious what others think! It occurs to me that I could also add `@since` tags, assuming the next version is `1.8` (this will require a major version change).

moodmosaic:
Thank you @tbidne! I'll take a close look. 🙏

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -118,46 +123,88 @@ runTreeN n m = do
...
-                if isFailure o then
-                  Just <$> loop (shrinks + 1) (n : revShrinkPath) o
-                else
-                  return Nothing
-
-          Right () ->
-            return OK
-  in
-    loop shrinks0 (reverse shrinkPath0)
+    runLoop $ loop shrinks0 (reverse shrinkPath0) initFailure initJournal initXs
```
moodmosaic, line 207: Look like it's reversed twice?

```hs
let initFailureReport =
      mkFailure
        shrinks0
        (ShrinkPath $ reverse shrinkPath0)   -- ← reversed here
```

down through:

```hs
runLoop $ loop shrinks0 (reverse shrinkPath0) initFailure initJournal initXs
                        -- ← seeded reversed too; loop then reverses again internally
```

hedgehog/src/Hedgehog/Internal/Runner.hs
```diff
@@ -597,3 +649,9 @@ checkParallel =
       , runnerVerbosity =
           Nothing
       }
+
+-- vendored from lifted-base
+timeout :: MonadBaseControl IO m => Int -> m a -> m (Maybe a)
```
moodmosaic, line 654: What was the reason to not `import System.Timeout.Lifted.timeout` from `lifted-base` instead?

hedgehog/src/Hedgehog/Internal/Property.hs
```diff
@@ -1267,6 +1285,15 @@ withShrinks :: ShrinkLimit -> Property -> Property
 withShrinks n =
   mapConfig $ \config -> config { propertyShrinkLimit = n }
 
+-- | Set the timeout -- in microseconds -- after which the test runner gives
```
moodmosaic: Timeout is best-effort, though. Perhaps we should document it.

hedgehog-example/src/Test/Example/Shrink.hs
```diff
@@ -0,0 +1,171 @@
...
+    startTime <- liftIO $ Clock.getMonotonicTime
+    annotateShow startTime
+    _ <- checkModPropGen delay30s (withShrinkTimeoutMicros 2000000)
+    endTime <- liftIO $ Clock.getMonotonicTime
+    annotateShow endTime
+    let timeElapsed = endTime - startTime
+    annotateShow timeElapsed
+    -- should be around 2
+    diff timeElapsed (>=) 1.5
+    diff timeElapsed (<=) 2.5
```
moodmosaic: These two can flake under CI CPU starvation/GC; and a non-allocating slow step might never be interrupted.

hedgehog/src/Hedgehog/Internal/Runner.hs
tbidne: Ah you are right. This is the final value, which ends up being reversed twice, hence the initial IORef should not be reversed at all.

Note that it actually doesn't matter here, since this is the initial call to `checkReport`, hence `shrinkPath0 == []`. Regardless, the reverse is unnecessary and misleading. Thanks for the catch! I added testing the `failureShrinkPath` to my example test.

hedgehog/src/Hedgehog/Internal/Runner.hs
tbidne: Saves an explicitly declared dependency, though of course `lifted-base` is a transitive dependency via `lifted-async`. I can declare it instead if you like.

hedgehog/src/Hedgehog/Internal/Property.hs
tbidne: What do you think of something like:

> This is based on base's [`timeout`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/System-Timeout.html#v:timeout), hence its caveats apply.

hedgehog-example/src/Test/Example/Shrink.hs
tbidne: Would you prefer this test be removed, then?

moodmosaic: I'd keep it, just I'd probably widen the upper bound: the regression it guards is "cancellation happens at ~2s instead of 30s", so diff timeElapsed (<=) 15 can be good enough.

hedgehog/src/Hedgehog/Internal/Property.hs
moodmosaic: Yep! Perhaps by having something like

```
-- | Set the timeout — in microseconds — after which the test runner gives up
--   on shrinking and reports the best counterexample found so far.
--
--   The timeout covers only the shrinking phase, not the search for the
--   original failure. Shrinking may also stop earlier if the 'ShrinkLimit'
--   is reached, see 'withShrinks'.
--
--   The timeout is enforced with 'System.Timeout.timeout', so it is
--   best-effort: it relies on an asynchronous exception and cannot interrupt
--   blocking foreign calls or non-allocating loops, and handlers that catch
--   asynchronous exceptions may delay it.
--
--   A timeout of @0@ disables shrinking entirely; negative values disable
--   the timeout.
```

(Best-effort docs, that is.)

hedgehog/src/Hedgehog/Internal/Runner.hs
moodmosaic: True! We can keep this as-is. 🙏

moodmosaic (commented):
LGTM! Just a few minor nits/comments and it should be ready to released as hedgehog-1.8 🎉

hedgehog/src/Hedgehog/Internal/Property.hs
tbidne: Looks good, thanks!

tbidne:
I believe all comments have been addressed.

hedgehog-example/src/Test/Example/Shrink.hs
```diff
@@ -0,0 +1,184 @@
...
+-- enough to not cause CI failures, and we cap the tests at 1 to keep the
+-- running time fast.
+prop_ShrinkTimeoutMicros :: Property
+prop_ShrinkTimeoutMicros =
+  withTests 1 . property $ do
+    -- Test generates [ 50 , 0 , 25 , 13 , 7 , 4 , 6 , 5 ]
+    -- The 1 s timeout combined with the 10 s delay on 13 means
+    -- shrinking will get stuck on 13, hence:
+    --   - only generate [50 , 0 , 25 , 13]
+    --   - final shrink value is 13
```
moodmosaic: `13` or `25`?

hedgehog-example/src/Test/Example/Shrink.hs
```diff
@@ -0,0 +1,184 @@
...
+    case reportStatus report of
+      Failed f -> do
+        1 === failureShrinks f
+        case failureAnnotations f of
+          [ann] -> do
+            "25" === failedValue ann
+            ShrinkPath [1] === failureShrinkPath f
+          _ -> failure
+      _ -> failure
+  where
```
moodmosaic: As per the [`STYLE_GUIDE.md`](https://github.com/hedgehogqa/haskell-hedgehog/blob/master/STYLE_GUIDE.md#do-not-use-where-only-let), we only use `let` instead of `where` (it's also in a few other places). However, don't let that bother you. This PR has been already a long journey and I can fix/adjust those myself. 🙏

moodmosaic (approved):
LGTM! Thank you for all the great work you’ve put into this! 🙏

I left a few final, non-blocking comments. Please don’t feel obliged to address them -- I can take care of them afterward if needed. I’ll leave the PR open for a day or so in case you’d like to make those tweaks; otherwise, very happy to merge it as-is.

hedgehog-example/src/Test/Example/Shrink.hs
tbidne: Ah good catch, thanks.

hedgehog-example/src/Test/Example/Shrink.hs
tbidne: Old habits die hard :slightly_smiling_face:.

moodmosaic (approved):

moodmosaic:
*Thank you*, @tbidne! 🎉

tbidne:
Thanks @moodmosaic!

moodmosaic:
We'll cut a release soon.

PR #489, opened 2023-05-25
Fix test skipping when tests have been discarded.

### LINKED ISSUE #487: recheckAt does not always reproduce a test failure
In hedgehog 1.2, a new feature `recheckAt` was added (#454) to skip directly to a failing test. However this sometimes fails: I have a property that fails, and says "This failure can be reproduced by running `recheckAt (Seed _ _) "..." <property>`", but running this command reports "gave up after 0 discards, passed 4 tests" (or, sometimes reports the test passed).

I have seen this on a few different tests in a real-life codebase, but certainly not all (i.e. it seems that there is a problem with only some properties). I have never seen anything similar on older releases of hedgehog.

I managed to minimize to the following generator and property (note that `recheckAt` fails to re-find a failure around 80% of the time):
```haskell
import Data.Maybe (catMaybes)
import Hedgehog (
  Gen,
  Property,
  discard,
  forAll,
  property,
  (===),
 )

tasty_replay_broken :: Property
tasty_replay_broken = property $ do
  tgt <- forAll $ genWTType KType
  src <- forAll $ genWTType KType
  case stripArgs tgt src of
    Just instTy -> do
      src === instTy
    _ -> discard

genWTType :: Kind -> Gen Type
genWTType k = do
  Gen.recursive Gen.choice [ehole] $ app : catMaybes [arrow]
  where
    ehole = pure TBase
    app = do TApp <$> genWTType (KFun KType k) <*> genWTType KType
    arrow =
      if k == KType
        then Just $ TFun <$> genWTType KType <*> genWTType KType
        else Nothing

data Type
  = TBase
  | TFun Type Type
  | TApp Type Type
  deriving (Eq, Show)

data Kind = KType | KFun Kind Kind
  deriving (Eq, Show)

stripArgs :: Type -> Type -> Maybe Type
stripArgs tgt ty =
  if tgt == ty
    then Just ty
    else case ty of
      TFun _ t -> stripArgs tgt t
      _ -> Nothing
```

A representative failure is
```
ghci> check tasty_replay_broken 
  ✗ <interactive> failed at src/Tests/Refine.hs:22:11
    after 7 tests and 13 discards.
    shrink path: 7:
  
       ┏━━ src/Tests/Refine.hs ━━━
    16 ┃ tasty_replay_broken :: Property
    17 ┃ tasty_replay_broken = property $ do
    18 ┃   tgt <- forAll $ genWTType KType
       ┃   │ TBase
    19 ┃   src <- forAll $ genWTType KType
       ┃   │ TFun TBase TBase
    20 ┃   case stripArgs tgt src of
    21 ┃     Just instTy -> do
    22 ┃       src === instTy
       ┃       ^^^^^^^^^^^^^^
       ┃       │ ━━━ Failed (- lhs) (+ rhs) ━━━
       ┃       │ - TFun TBase TBase
       ┃       │ + TBase
    23 ┃     _ -> discard
  
    This failure can be reproduced by running:
    > recheckAt (Seed 3098031441793650136 15775858570470438977) "7:" <property>
  
False
ghci> recheckAt (Seed 3098031441793650136 15775858570470438977) "7:" tasty_replay_broken 
  ⚐ <interactive> gave up after 0 discards, passed 7 tests.
```
Note that sometimes (but more rarely) `recheckAt` will show the test passed.

ChickenProp:
Closes #487.

In #454 we introduced test skipping, with the idea that a failed test could report a way for you to jump back to reproduce it without all the preceding tests.

But it didn't work if any of the preceding tests had been discarded, because each discard also changes the seed and the size. Users could manually add the discard count to the test count in the `Skip`, but that's no fun. Plus, it wouldn't work if the test count plus discard count exceeded the test limit, because that would generate a success without running any tests.

So now a `Skip` (other than `SkipNothing`) includes a `DiscardCount` as well as a `TestCount`. It's rendered in the compressed path as `testCount/discardCount`, or just `testCount` if `discardCount` is 0. The exact sequence of passing tests and discards doesn't affect the final seed or size, so the counts are all we need.

This changes an exposed type, so PVP requires a major version bump.

This is awkward for me personally because it's incompatible with the approach I took in #474, I'll leave a comment there with more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Version 1.4 (unreleased)
+
+* Fix skipping to tests/shrinks when tests have been discarded ([#489][489], [@ChickenProp][ChickenProp])
+
 ## Version 1.3 (2023-06-22)
 
 * Better documentation for `Var` ([#491][491], [@endgame][endgame])
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -312,7 +312,7 @@ newtype TestCount =
 --
 newtype DiscardCount =
   DiscardCount Int
-  deriving (Eq, Ord, Show, Num, Enum, Real, Integral)
+  deriving (Eq, Ord, Show, Num, Enum, Real, Integral, Lift)
 
 -- | The number of discards to allow before giving up.
 --
@@ -355,7 +355,10 @@ data Skip =
   -- | Skip to a specific test number. If it fails, shrink as normal. If it
   --   passes, move on to the next test. Coverage checks are disabled.
   --
-  | SkipToTest TestCount
+  --   We also need to count discards, since failing "after 7 tests" points at a
+  --   different generated value than failing "after 7 tests and 5 discards".
+  --
+  | SkipToTest TestCount DiscardCount
 
   -- | Skip to a specific test number and shrink state. If it fails, stop
   --   without shrinking further. If it passes, the property will pass without
@@ -365,7 +368,7 @@ data Skip =
   --   the direct path from the original test input to the target state - will
   --   be tested too, and their results discarded.
   --
-  | SkipToShrink TestCount ShrinkPath
+  | SkipToShrink TestCount DiscardCount ShrinkPath
   deriving (Eq, Ord, Show, Lift)
 
 -- | We use this instance to support usage like
@@ -402,13 +405,17 @@ newtype ShrinkPath =
 --   roughly interpret it by eyeball.
 --
 skipCompress :: Skip -> String
-skipCompress = \case
-  SkipNothing ->
-    ""
-  SkipToTest (TestCount n) ->
-    show n
-  SkipToShrink (TestCount n) sp ->
-    show n ++ ":" ++ shrinkPathCompress sp
+skipCompress =
+  let
+    showTD (TestCount t) (DiscardCount d) =
+      show t ++ (if d == 0 then "" else "/" ++ show d)
+  in \case
+    SkipNothing ->
+      ""
+    SkipToTest t d->
+      showTD t d
+    SkipToShrink t d sp ->
+      showTD t d ++ ":" ++ shrinkPathCompress sp
 
 -- | Compress a 'ShrinkPath' into a hopefully-short alphanumeric string.
 --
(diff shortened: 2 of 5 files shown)


TysonMN:
Why is it necessary to track `testCount` and `discardCount` separately? I was expecting that it would suffice to only track their sum.

TysonMN:
I think that message is not helpful when recording

ChickenProp:
I agree it's not really helpful, but if it's going to be there it seems kinda confusing if it doesn't match the one they got before. We could remove it when replaying I guess, but that would involve more changes.

Ultimately, it would be possible to only track (sum of tests and discards), but I think tracking them separately is mildly more user friendly and mildly easier to implement.

moodmosaic (commented):
@ChickenProp, could this break any external integration?

moodmosaic:
For the rest of the changes, as also mentioned in #454, it'd be nice also here if our friends can take a look wrt tasty/hspec hedgehog integrations. /cc @parsonsmatt @dalaing @gwils

gwils:
@moodmosaic This is easy for us to support; I have a patch ready to go. It will be especially easy if you do release this as part of hedgehog 1.3 instead of 1.2.x, since then our upper bounds will work correctly without revision.

moodmosaic:
Sure, no problem for 1.4 as it's also more correct.

If that works, we can merge after @ChickenProp resolves the conflicts and then do a hedgehog 1.4 release this weekend. 🦔

ChickenProp:
> could this break any external integration?

I might be misinterpreting, but is the question "could this break packages that depend on hedgehog?"

I don't expect downstream users to commit code with a `ShrinkPath`, so it shouldn't cause problems for them. Packages integrating with hedgehog might need changes. But I think hspec-hedgehog didn't need changes to support v1.2 when we got shrink paths, so I don't expect it to need changes for this; and it sounds like tasty is on top of things.

moodmosaic:
Thank you, @ChickenProp 🚀 A `hedgehog 1.4` release is planned for this weekend. 🦔

moodmosaic:
Release just went out. 🦔
* https://hackage.haskell.org/package/hedgehog-1.4
* https://hackage.haskell.org/package/hedgehog-quickcheck-0.1.1 ([r6](https://hackage.haskell.org/package/hedgehog-quickcheck-0.1.1/revisions/))

PR #491, opened 2023-06-08
State: Remove HTraversable from haddock

endgame:
Looks like this comment was missed in the `HTraversable`->`barbies` migration.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/State.hs
+++ b/hedgehog/src/Hedgehog/Internal/State.hs
@@ -170,8 +170,8 @@ instance Ord1 Concrete where
 --   type of variable because it is used in both the generation and the
 --   execution phase.
 --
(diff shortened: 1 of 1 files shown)


hedgehog/src/Hedgehog/Internal/State.hs
```diff
@@ -170,8 +170,8 @@ instance Ord1 Concrete where
 --   type of variable because it is used in both the generation and the
 --   execution phase.
 --
---   The order of arguments makes 'Var' 'HTraverable', which is how 'Symbolic'
---   values are turned into 'Concrete' ones.
+--   The order of arguments makes 'Var' 'FunctorB' and 'TraversableB', which is
```
moodmosaic, line 173: Good catch 👍

FWIW, the actual replace happened in https://github.com/hedgehogqa/haskell-hedgehog/pull/412.

moodmosaic (approved):

Issue #459, opened 2022-06-09
Return multiple Vars from a single Command


ChickenProp:
I have a state machine where one Command's output type is `(FooId, BarId)`. That gets put into the model, so I have `ids :: Var (FooId, BarId) v` in my state. Then any time I want to pass either `FooId` or `BarId` into a `commandExecute`, it needs to get both of them and deconstruct.

I think it would be possible to change things so each Command can return multiple Vars at once. The state would get `fooId :: Var FooId v, barId :: Var BarId v`, the same as if they'd come from different Commands.

The sketch is that `output` would move from being kind `Type` to `(Type -> Type) -> Type` like `input`, parameterized by `Symbolic` or `Concrete`. `commandExecute` returns `m (output Concrete)`. `Update` gets `output v` instead of `Var output v`, and `Ensure` gets `output Concrete` instead of just `output`. Where I currently return `(fooId, barId)`, I'd instead return `Tuple2B (Concrete fooId, Concrete barId)`, where

```haskell
data Tuple2B a b f = Tuple2B (f a) (f b) -- feels like this should already exist somewhere?
```

Then we need some instances from [barbies](https://hackage.haskell.org/package/barbies-2.0.3.1/docs/Data-Functor-Barbie.html) to generate the vars. From a bit of exploration, I think giving `Tuple2B a b` instances of `ApplicativeB` and `ConstraintsB` is enough. `bpure` lets us generate `Tuple2B (Const ()) (Const ()) :: Tuple2B a b (Const ())`, and then we can `btraverseC` to generate a `Name` for each of those `Const`s. (We need `ConstraintsB` because of the `Typeable` constraint in `Symbolic`.)

This does make single-var Commands a bit more complicated, they now return `Var (Concrete x)` instead of just `x`. (Or I think we'd want a new `newtype SoloB a f = SoloB (f a)` on the grounds that `Var` feels like it only works here "by coincidence". I'm a bit surprised there's nothing like `SoloB` in the barbies package.) Commands currently returning `()` can just return barbies' `Unit`. I think it would be possible to do this mostly backwards-compatible, similar to how #456 does it, though perhaps more awkward.

I've made a bit of progress on this that makes me think it works, but not yet in a sharable state. If it's wanted I can finish it off.

ocharles:
This is an interesting idea. My usually approach for returning multiple values is to work within the framework that Hedgehog gives us, but to then essentially store defunctionalized functions to extract elements where I need them.

[...]

PR #493, opened 2023-06-16
Hedgehog.Internal.State: Add instance Applicative Concrete

endgame:
Make `Concrete` an instance of `Applicative`. This is useful if one wants to wrap `Symbolic` values in a free `Applicative` and "pretend" to compute on `Symbolic` values during an update hook. Example: package `free` provides [`retractAp :: Applicative f => Ap f a -> f a`](https://hackage.haskell.org/package/free-5.2/docs/Control-Applicative-Free.html#v:retractAp).

Inspired by [this discussion](https://github.com/hedgehogqa/haskell-hedgehog/issues/459#issuecomment-1594171963)


moodmosaic:
Thank you, @endgame 👍 Would it be possible to add an example in tests?

PR #499, opened 2023-07-11
Wrap `evalMaybeM` and `evalEitherM` in `withFrozenCallStack`

### LINKED ISSUE #497: `evalMaybeM` doesn't produce the correct callstack
If we use `evalMaybeM` we get `the value was Nothing`, but it's not attached to the line calling `evalMaybeM`. I think `evalMaybeM` needs a `withFrozenCallStack`, but I'm not quite sure where - probably around the whole definition?

ocharles:
Fixes #497

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Property.hs
+++ b/hedgehog/src/Hedgehog/Internal/Property.hs
@@ -1016,7 +1016,7 @@ evalEither = \case
 --
 evalEitherM :: (MonadTest m, Show x, MonadCatch m, HasCallStack) => m (Either x a) -> m a
 evalEitherM =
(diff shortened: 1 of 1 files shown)


moodmosaic:
LGTM 👍 Do you think we could add some (reliable, trustworthy) test(s) for this change?

moodmosaic (commented):
LGTM 👍 Do you think we could add some (reliable, trustworthy) test(s) for this change?

ocharles:
I don't think I'm likely to find the time to do anymore work on this, sorry.

moodmosaic:
No worries 👍

PR #506, opened 2023-11-12
Exposes more state machine functionality.

dalaing:
This allows us to do fancier generation than we can do with just `sequential`.

I'm using this to allow for explicitly generating actions over phases - the first phase sets up some account, the second phases uses the accounts to add data to the application under test, and the third phase is reading the data.

(I'm mostly interested in using this for benchmarking, by fixing a seed and size and having some setup phases run before the actual benchmark phases)

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/State.hs
+++ b/hedgehog/src/Hedgehog/Internal/State.hs
@@ -38,13 +38,19 @@ module Hedgehog.Internal.State (
   , commandGenOK
 
   -- * Actions
+  , Context(..)
   , Action(..)
   , Sequential(..)
   , Parallel(..)
   , takeVariables
   , variablesOK
+  , mkContext
+  , contextUpdate
+  , contextNewVar
   , dropInvalid
   , action
+  , genActions
+  , executeUpdateEnsure
   , sequential
   , parallel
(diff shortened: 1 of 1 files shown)


moodmosaic (commented):
LGTM 👍

Can we include some example/test showing what’s possible? That’d be great to have, so the users know how to use the stuff we’re exporting.

dalaing:
For sure. I'll try to get some examples together in the next week or so.

dalaing:
It turns out I needed more constraints inside of the Command constructor, and had to copy-and-paste huge chunks of State.hs to get my benchmarking to go. Should I close this? Or leave it in case others might need it? I can always recreate this and open a new PR if I end up with a new concrete use case for it.

moodmosaic:
Thank you for your contribution! We appreciate your effort. 🙏 💯 

Regarding the PR, since it sounds like the changes might not be needed in the current scope of the project, and you're also considering different approaches, I think it would be best to close this PR for now. 

Please do feel free to reopen it or submit a new PR if you find that these changes become relevant in the future or if you have other enhancements to suggest.

Thanks again for your willingness to contribute, and looking forward to any future contribution. 🚀

PR #510, opened 2024-02-03
Bump barbies upper bound

jchia:


moodmosaic:
Thank you! 👍 

* Changed the library component's library dependency on `barbies` from

  ```
  >=1.0 && <2.1
  ```
  to
  ```
  >=1.0 && <2.2
  ```

Issue #512, opened 2024-03-12
Relax upper bound on `time` to `< 1.15`


tomjaguarpaw:
This succeeds, so I guess a Hackage revision should relax the bound on `time` to `< 1.15` (currently it's `< 1.13`).

* `cabal repl -z -b hedgehog -b 'time>=1.14' --allow-newer=time`

moodmosaic:
Thank you! If you open a PR, we can merge it and do a hackage revision right after.

tomjaguarpaw:
OK: https://github.com/hedgehogqa/haskell-hedgehog/pull/513

PR #513, opened 2024-03-13
Relax time dependency to < 1.15

tomjaguarpaw:
Fixes https://github.com/hedgehogqa/haskell-hedgehog/issues/512


moodmosaic:
Thank you! 👍 

* Changed the library component's library dependency on `time` from

  ```
  >=1.4 && <1.13
  ```
  to
  ```
  >=1.4 && <1.15
  ```

Issue #518, opened 2024-04-30
Site is down


jezen:
The https://hedgehog.qa/ site returns a 404.

moodmosaic:
@jezen, thank you for reporting this. @ploeh [reported](https://twitter.com/ploeh/status/1783549441120571446) this on X too. Given https://github.com/hedgehogqa/haskell-hedgehog/commit/114c1e52b479dc2795c42b655f73a15fd26d747d, it might take me some time to figure out if we can get the website up.

jezen:
I am very sorry to see that news.

Let us know if there is some way the community can help.

PR #525, opened 2024-05-10
temorary solution for #518

aaaaargZombies:
Hello, I'm new to Haskell and exploring the ecosystem. Sorry to see the site for this project was down and the reasons why.

A suggested temporary solution to #518 that still points people to the written resources and recommendations for talks.

I understand this is more than a practical issue for those involved in the project so if this is not the way you'd like to go just close the PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ https://preview.webflow.com/preview/hedgehogqa?utm_medium=preview_link&utm_sourc
 
 <div align="left">
 
-[Hedgehog](http://hedgehog.qa/) automatically generates a comprehensive array of test cases, exercising your software in ways human testers would never imagine.
+[Hedgehog](https://web.archive.org/web/20240329171125/https://hedgehog.qa/) automatically generates a comprehensive array of test cases, exercising your software in ways human testers would never imagine.
 
 Generate hundreds of test cases automatically, exposing even the most insidious of corner cases. Failures are automatically simplified, giving developers coherent, intelligible error messages.
 


moodmosaic (approved):
Thank you! That’s an interesting idea and definitely better than a 404. Probably we can merge this, as a temporary fix.

Perhaps a long term solution can be hosting the contents of hedgehog.qa website on GitHub Pages. This assumes that there’s a way to reliably (and in full) fetch the content from web.archive.org, which I’ve never tried.

moodmosaic (changes requested):
Thanks for the PR. I'm reconsidering the use of a web archive link, as it might not be ideal for an active project.

As an alternative, we could temporarily remove the hedgehog.qa link from the README.md entirely. We could handle this in two steps in our git history:

1. Replace the current link with the web archive version. (This PR)
2. Remove the link altogether. (Can be in this PR, too.)

This method allows us to keep a record of the changes and potentially add a new link later when we transition to GitHub Pages.

I would appreciate your thoughts and any other suggestions on this approach.

cc: @sol @ocharles @ysangkok @ChickenProp @endgame @ploeh

sol:
Migrating to GitHub pages sounds good to me: https://github.com/hedgehogqa/haskell-hedgehog/issues/526

ChickenProp:
I think github pages is the ideal solution, but an archive link is better than no link or a broken link. It definitely looks weird for an active project, but a brief footnote explaining it would make it less weird.

PR #529, opened 2024-05-14
hedgehog-{example,quickcheck,test-laws}: Raise QuickCheck bound

JackKelly-Bellroy:
We have an `allow-newer: hedgehog-quickcheck:QuickCheck` working internally, and this PR passed `cabal test all --constraint 'QuickCheck ^>=2.15'`.

Can we please also get metadata revisions on Hackage?


moodmosaic:
Thank you, @JackKelly-Bellroy. Sounds good 👍 We can do a metadata revision. I’ll post here once it’s done.

Issue #522, opened 2024-05-04
Allow `hspec-hedgehog` to produce more compact output


sol:
This requires:
- [x] #519
- [x] #520
- [x] #523
- [x] #524

Before:
![before](https://github.com/hedgehogqa/haskell-hedgehog/assets/461132/315b38ea-8c80-4ee7-87cc-57a896415c36)

After:
![after](https://github.com/hedgehogqa/haskell-hedgehog/assets/461132/5984c2e6-6bd8-4fdf-8afe-c6fbd5c87ffe)

sol:
@moodmosaic that's all that I have on my wish list.  After that, I'm pretty happy with the output.

moodmosaic:
> @moodmosaic that's all that I have on my wish list. After that, I'm pretty happy with the output.

Planning for a release after #525. 🎉

sol:
https://github.com/hedgehogqa/haskell-hedgehog/pull/519#issuecomment-2105614330

@ChickenProp e could still include the "shrink path", I guess.

From what I understand, this only affects performance. Do you regularly run into situations where this makes a practical difference?

ChickenProp:
I regularly run into situations where it makes a big practical difference to performance :p Like, it can be several minutes versus a few seconds.

But also, it can really help debugging. I often use print or trace statements to help, and if I don't use the shrink path, the last lines of output are often from tests that passed. Or I'll do stuff with a database, and then the database state at the end has been changed by tests that passed. With the shrink path, I know the last attempt failed, which is super helpful. (If I want that with quickcheck, I kind of have to binary search to find the lowest `--qc-max-shrinks` value that gives the full number of successful shrinks.)

(I did notice that the shrink path is also visible at the top of the failure message, so there's that too. But it's a lot easier for me if it's at the bottom.)

sol:
> I did notice that the shrink path is also visible at the top of the failure message, so there's that too.

It's gone now, see screens, but we could bring it back.

 > But it's a lot easier for me if it's at the bottom.

Given that the output will be much more compact now (mostly due to https://github.com/hedgehogqa/haskell-hedgehog/pull/505), does this still matter?

ChickenProp:
> It's gone now, see screens, but we could bring it back.

[...]

PR #530, opened 2024-05-20
Make it build with ghc 9.10

erikd:
Tested on Linux with GHC versions `8.10.7`, `9.6.5`, `9.8.2` and `9.10.1`

### DIFF
--- a/hedgehog-test-laws/hedgehog-test-laws.cabal
+++ b/hedgehog-test-laws/hedgehog-test-laws.cabal
@@ -54,13 +54,12 @@ test-suite test
   build-depends:
       hedgehog
     , base                            >= 3          && < 5
-      -- https://github.com/conal/checkers/issues/44
-    , checkers                        >= 0.5        && < 0.5.5
+    , checkers                        >= 0.5        && < 0.8
     , QuickCheck                      >= 2.10       && < 2.16
-    , tasty                           >= 1.2        && < 1.3
+    , tasty                           >= 1.2        && < 1.6
     , tasty-expected-failure          >= 0.11       && < 0.12
     , tasty-quickcheck                >= 0.10       && < 0.11
-    , transformers                    >= 0.5        && < 0.6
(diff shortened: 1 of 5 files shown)


hedgehog-test-laws/hedgehog-test-laws.cabal
```diff
@@ -54,13 +54,12 @@ test-suite test
   build-depends:
       hedgehog
     , base                            >= 3          && < 5
-      -- https://github.com/conal/checkers/issues/44
-    , checkers                        >= 0.5        && < 0.5.5
+    , checkers                        >= 0.5        && < 0.8
```
erikd, line 57: The checkers issues has been fixed and I assume the new version on Hackage has the fix.

moodmosaic: Latest version on hackage is `0.6` so `< 0.8` will keep CI happy for a while.

moodmosaic (approved):
Thank you, @erikd 💯

erikd:
A new release would be much appreciated too.

moodmosaic:
Let's wait to see what we will end up doing in https://github.com/hedgehogqa/haskell-hedgehog/issues/522#issuecomment-2119766579 and then we can release. 🎉

tomjaguarpaw:
I wonder if we could at least get a Hackage revision to 1.4, to loosen these bounds?

moodmosaic:
Absolutely! Done 👍 

---

* Changed the library component's library dependency on `template-haskell` from

  ```
  >=2.10 && <2.22
  ```
  to
  ```
  >=2.10 && <2.23
  ```

And hopefully a release coming up this week. 🎉

tomjaguarpaw:
Thank you! That let me test 9.10 on tomland: https://github.com/kowainik/tomland/pull/445

moodmosaic:
Awesome 🎉

PR #534, opened 2024-07-24
Release 1.5

moodmosaic:
@sol, @endgame, @erikd, @JackKelly-Bellroy, @jchia, @moodmosaic, @ocharles, @tomjaguarpaw, @Vekhir

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,23 @@
-## Version 1.5 (unreleased)
+## Version 1.5 (2024-07-25)
 
-* Remove redundant `Show` constraints on `evalMaybe`, `evalMaybeM`
+* Bump containers and filepath dependencies ([#533](533), [@erikd][erikd])
+* Hedgehog.Internal.Report: Add `configPrintShrinkPath` ([#531](531), [@sol][sol])
+* Allow GHC 9.10 ([#530](530), [@erikd][erikd])
+* hedgehog-{example,quickcheck,test-laws}: Raise QuickCheck bound ([#529](529), [@JackKelly-Bellroy][JackKelly-Bellroy])
+* Allow to omit icons at the start of important output lines ([#524](524), [@sol][sol])
+* Allow to omit source location from report ([#523](523), [@sol][sol])
+* Fix "insufficient coverage" message ([#521](521), [@sol][sol])
+* Support empty property names ([#520](520), [@sol][sol])
+* Allow to disable the `recheckAt`-message ([#519](519), [@sol][sol])
+* Remove redundant `Show` constraints: `evalMaybe`, `evalMaybeM` ([#516](516), [@endgame][endgame])
+* Relax upper bound on `ansi-terminal` ([#515](515), [@moodmosaic][moodmosaic])
+* Report.hs refactoring ([#514](514), [@sol][sol])
+* Relax time dependency to < 1.15 ([#513](513), [@tomjaguarpaw][tomjaguarpaw])
+* Bump barbies upper bound ([#510](510), [@jchia][jchia])
+* Support primitive 0.9 ([#509](509), [@Vekhir][Vekhir])
+* Show less context on failure ([#505](505), [@sol][sol])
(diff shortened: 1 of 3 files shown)


moodmosaic:
hedgehog-1.5 package [candidate](https://hackage.haskell.org/package/hedgehog-1.5/candidate). If all looks good to you, we can do the actual release any time.

endgame:
While I haven't looked at all the details, the mass CI failures seem concerning.

Vekhir:
The CI failures afaict result from the fact that `hedgehog-quickcheck` doesn't support `hedgehog 1.5`. Probably needs a version bump [here](https://github.com/hedgehogqa/haskell-hedgehog/blob/328a07f509ce1ae4c2996bfd67dbd1a545bb3b88/hedgehog-quickcheck/hedgehog-quickcheck.cabal#L53) and a corresponding revision.

moodmosaic:
@Vekhir, good catch, fixed and force-pushed. 👍

Vekhir:
Right, it's markdown.
That's all from me.

moodmosaic:
@Vekhir, thanks, it has been fixed in both markdown and package [candidate](https://hackage.haskell.org/package/hedgehog-1.5/candidate). Can you double-check?

Vekhir:
Yep, has been updated in the candidate.

moodmosaic:
>`hedgehog-quickcheck` doesn't support `hedgehog 1.5`. Probably needs a version bump

Published new [revision](https://hackage.haskell.org/package/hedgehog-quickcheck-0.1.1/revisions/) for [hedgehog-quickcheck-0.1.1](https://hackage.haskell.org/package/hedgehog-quickcheck-0.1.1):

* Changed the library component's library dependency on `hedgehog` from

  ```
  >=0.5 && <1.5
  ```
  to
  ```
  >=0.5 && <1.6
  ```

moodmosaic:
All comments addressed, all checks green. Shall we publish `hedgehog-1.5`? 👀

moodmosaic:
https://hackage.haskell.org/package/hedgehog-1.5 🎉

mitchellwrosen:
Hello all, just curious, are there any breaking changes in this release that motivated the major version bump? (I should also ask: does this package follow the PVP?)

Looking over the changelog, I'm not seeing any, but it's a little bit hard to tell. I got pinged by the Stackage maintainers to update the bounds in a few of my test suites, which I should only have to do if there has been a backwards-incompatible change. If not, in the future, a minor version bump under such circumstances would be greatly appreciated!

endgame:
I'm speculating here, but it seems like the [original author passed away](https://github.com/hedgehogqa/haskell-hedgehog/commit/114c1e52b479dc2795c42b655f73a15fd26d747d) not that long ago, and the new maintainers may be getting up to speed with things. A major version bump might have been the safest option to get things moving.

moodmosaic:
Thanks for reaching out. This was a complete oversight on my part, and I apologize.

mitchellwrosen:
No worries @moodmosaic!

sol:
@mitchellwrosen @moodmosaic here is an other perspective:

1. A breaking change is not only when the type of a function changes, but also when its *behavior* changes. [^1]
2. For `IO` actions, the behavior also includes things like output to `stdout`.
3. Programs are `IO` actions, and by extension the output of a program is part of its behavior.

Both #499 and #521 change the output of failing tests, and by extension not only change the behavior of programs that are constructed with `hedgehog` but also the behavior of exported `IO` actions and functions.

Now you could always label something as a "bug fix" and only do a minor or even patch level bump. But you can also acknowledge the fact that some people may rely on the output format and instead do a major version bump.  There is no fixed rule here!

I don't want to offer specific recommendations, but I hope that we at least agree that the idea that "everything still  compiles", so it's "not a breaking change" is inherently flawed.

[^1]: Mathematically, the type of a function only describes the *domain* and *codomain*, while the behavior of a function is akin to its *image*.

moodmosaic:
Thank you for sharing this insightful perspective on the nature of breaking changes, particularly regarding behavior and output in `IO` actions. I've sensed this too but never expressed it as well as you did. 💯

PR #538, opened 2024-08-23
Add callstacks to generators that can error.

ChickenProp:
Previously, `val <- Gen.element []` would give an error like

    Hedgehog.Gen.element: used with empty Foldable
    CallStack (from HasCallStack):
      error, called at src/Hedgehog/Internal/Gen.hs:1197:5 in hedgehog-1.4-FEcjASqhnyiHkb8BJanjYM:Hedgehog.Internal.Gen

Which isn't very helpful. Now the callstack entry points to my own code, instead.

I guess this will need a version bump since it changes types?

I think I got all the generators that are expected to throw errors. Some call `error` but describe it as internal, and I didn't touch those.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,11 +14,9 @@ jobs:
       matrix:
         os:    [macos-latest, ubuntu-latest, windows-latest]
         cabal: ["3.10.1.0"]
-        ghc:   ["8.0.2", "8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.7", "9.4.4", "9.8.2", "9.10.1"]
+        ghc:   ["8.2.2", "8.4.4", "8.6.5", "8.8.4", "8.10.7", "9.2.7", "9.4.4", "9.8.2", "9.10.1"]
         exclude:
           # https://github.com/haskell/text/pull/404
-          - os: windows-latest
-            ghc: "8.0.2"
           - os: windows-latest
             ghc: "8.2.2"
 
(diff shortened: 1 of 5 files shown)


ChickenProp:
Welp, apparently this fails on 8.0.2 :/

I imagine I can fix that somehow (we use `withFrozenCallStack` elsewhere, so it must be workable). But before I do, to check: do we still want support going that far back? 8.2.1 [was released](https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history) in 2017.

moodmosaic (commented):
Thank you for the PR!

The addition of `HasCallStack` and `withFrozenCallStack` looks good wrt improving error diagnostics. However, since this change modifies the function signatures, it could potentially break backward compatibility for users with explicit type annotations. So we’ll need to consider a major version bump.

Regarding GHC 8.0.2, I agree that it’s reasonable to drop support, allowing us to focus on more recent releases.

ChickenProp:
Yeah, agree it'll need a major version bump.

I can put either or both of the version bump and dropping 8.0.2 support in this PR if you want, or they can be separate.

moodmosaic:
Could be done in the same PR, just separate Git commit. And then we may update the PR title.

ChickenProp:
Okay, done both of those.

moodmosaic (approved):
@ChickenProp, thank you! The changes look good, so I’m happy to approve.

I’ll leave some room for other contributors to review, and if we don’t hear back, we can merge later this week. Hope that's okay! Thanks again for your work on this.

PR #546, opened 2025-05-19
Give a better implementation for `(*>)`

L0neGamer:
~~The default implementation of `(*>)` uses `(<*>)` which tries to perform each branch in parallel. However, since we only get the value from the right argument, we should just return that argument.

I don't implement `(>>)` similarly due to my attempts with https://github.com/haskell/core-libraries-committee/issues/333, but if that proposal falls through I'll be back.~~

Simply cleanup some instance definitions and add a comment.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -499,6 +499,9 @@ instance Functor m => Functor (GenT m) where
 --
 -- implementation: parallel shrinking
 --
+-- | This Applicative instance is not lawful with regards to the Monad instance.
+-- This is because using applicative allows us to do parallel shrinking, but
+-- Monad does not allow that.
 instance Monad m => Applicative (GenT m) where
   pure =
     fromTreeMaybeT . pure
--- a/hedgehog/src/Hedgehog/Internal/Gen.hs
+++ b/hedgehog/src/Hedgehog/Internal/Gen.hs
@@ -528,9 +528,6 @@ instance Monad m => Applicative (GenT m) where
 --          runGenT size sm m
 
 instance Monad m => Monad (GenT m) where
-  return =
-    pure
-
   (>>=) m k =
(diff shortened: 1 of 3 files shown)


ocharles:
This doesn't look sound - what if I'm using `GenT State`? You're not going to run the left potentially-stateful computation at all!

L0neGamer:
Damn, you've a point... I'll have another go

L0neGamer:
@ocharles if you don't mind, why is mzip and (<*>) different for TreeT?

I'm not sure why the below was or wasn't implemented
```diff
diff --git a/hedgehog/src/Hedgehog/Internal/Tree.hs b/hedgehog/src/Hedgehog/Internal/Tree.hs
index 0970ec1..d95aa62 100644
--- a/hedgehog/src/Hedgehog/Internal/Tree.hs
+++ b/hedgehog/src/Hedgehog/Internal/Tree.hs
@@ -428,9 +428,9 @@ instance (Eq1 m, Eq a) => Eq (TreeT m a) where
 instance Applicative m => Applicative (NodeT m) where
   pure x =
     NodeT x []
-  (<*>) (NodeT ab tabs) na@(NodeT a tas) =
-    NodeT (ab a) $
-      map (<*> (fromNodeT na)) tabs ++ map (fmap ab) tas
+  (<*>) nf@(NodeT f tfs) na@(NodeT a tas) =
+    NodeT (f a) $
+      map (<*> fromNodeT na) tfs ++ map (fromNodeT nf <*>) tas
 
 instance Applicative m => Applicative (TreeT m) where
   pure =
```

ocharles:
Sorry, I don't know why they are different - I'm not familiar with the underlying code.

moodmosaic:
I think this makes `(*>)` behavior consistent with `do` notation, where you'd expect sequential execution.

L0neGamer:
I've changed this to instead just cleanup some instances and add a comment. I can adjust the changelog if wanted, but this seems minor.

PR #554, opened 2025-09-09
Remove `fromJust` call by using a separate existential wrapper

HuwCampbell:
This is just a tidy up to remove calls to `error`.


hedgehog/src/Hedgehog/Internal/State.hs
```diff
@@ -402,11 +401,26 @@ data Command gen m (state :: (Type -> Type) -> Type) =
         [Callback input output state]
     }
 
--- | Checks that input for a command can be executed in the given state.
+
+-- | An actionable command is a command that can be turned into
+--   an action by running its generator.
+--
```
moodmosaic, line 407: Perhaps add a note that we only construct actions where `inputGen state0` was `Just`, matching the new invariant.

hedgehog/src/Hedgehog/Internal/State.hs
```diff
@@ -35,7 +35,6 @@ module Hedgehog.Internal.State (
   -- * Commands
   , Command(..)
   , Callback(..)
-  , commandGenOK
```
moodmosaic, line 38: Note to self: Remember to bump the version appropriately for this.

moodmosaic (commented):
Nice! No second `inputGen state0` for the chosen command. 🎉

hedgehog/src/Hedgehog/Internal/State.hs
moodmosaic: See https://github.com/hedgehogqa/haskell-hedgehog/pull/554#issuecomment-3287624883.

Issue #555, opened 2025-09-10
Doesn't build under 9.0?


tomjaguarpaw:
It seems that `hedgehog-1.6` doesn't build with 9.0, perhaps something to do with simplified subsumption?

```
src/Hedgehog/Internal/Gen.hs:1211:12: error:
Error:     • Couldn't match type: HasCallStack => m a
                     with: f0 a
      Expected: f0 a -> m a
        Actual: (HasCallStack => m a) -> m a
    • In the first argument of ‘(.)’, namely ‘withFrozenCallStack’
Error: [Cabal-7125]
Failed to build hedgehog-1.6. See the build log above for details.

      In the expression:
        withFrozenCallStack
          . \case
              [] -> error "Hedgehog.Gen.element: used with empty list"
              xs
                -> do n <- integral_ $ Range.constant 0 (length xs - 1)
                      return $ xs !! n
      In an equation for ‘element_’:
          element_
            = withFrozenCallStack
                . \case
                    [] -> error "Hedgehog.Gen.element: used with empty list"
                    xs
                      -> do n <- integral_ $ Range.constant 0 (length xs - 1)
                            ....
    • Relevant bindings include
        element_ :: [a] -> m a
          (bound at src/Hedgehog/Internal/Gen.hs:1211:1)
     |
1211 | element_ = withFrozenCallStack . \case
     |            ^^^^^^^^^^^^^^^^^^^
```

https://github.com/kowainik/relude/actions/runs/17612390601/job/50037135832?pr=479#step:7:423

moodmosaic:
Hmm, perhaps we can just expand the definition to make the callstack constraint explicit.

PR #557, opened 2025-09-10
eta expand so it works on GHC 9.0; test on GHC 9.0.2

tomjaguarpaw:
where there is simplified subsumption

Fixes https://github.com/hedgehogqa/haskell-hedgehog/issues/555


moodmosaic (approved):

tomjaguarpaw:
(In fact I think `1.6.0.1` would be a valid version bump, since no external behaviour changes.)

moodmosaic:
@tomjaguarpaw, yes, that’s the plan. @bowtiedeadone works in #559 which will help. In the meantime we can release early this week. Hope it helps. Thank you!

PR #559, opened 2025-09-11
Add automated release workflow

BowTiedRadone:
This PR adds a GitHub Actions workflow to automate the release process for the Hedgehog library. Key features:

- Builds and packages the Hedgehog library using the latest `GHC` and `cabal-install` via `GHCup`.
- Uses `cabal sdist` to generate the tarball matching maintainer/Hackage requirements.
- Checks the package with `cabal check` before upload.
- Uploads the tarball to Hackage using a repository secret (`HACKAGE_TOKEN`).
- Only triggers on new tags.

**Notes**

The workflow expects a valid `HACKAGE_TOKEN` secret to be set in the repository for Hackage uploads. The tarball creation logic is designed to match the official Hackage release process as closely as possible.

**Considerations**

The new workflow's output was tested locally using `act` and diff-checked agains the latest official Hackage release. This PR should be merged after one of the maintainers can confirm the `HACKAGE_TOKEN` secret addition.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-24.04
 
     steps:
-      - name: Checkout repository with full history
+      - name: Checkout repository
         uses: actions/checkout@v5
         with:
           fetch-depth: 1


.github/workflows/release.yml
```diff
@@ -0,0 +1,58 @@
...
+    tags:
+      - "*"
+
+jobs:
+  release:
+    name: Build and Release Hedgehog
+    runs-on: ubuntu-24.04
+
+    steps:
+      - name: Checkout repository with full history
```
moodmosaic, line 14:
```suggestion
      - name: Checkout repository
```

moodmosaic (commented):
LGTM. Thank you for your contribution! I’m curious to know what @vrom911 and/or @chshersh think (if their time permits). 🙏

moodmosaic:
@BowTiedRadone, what would be the easiest way to test this?

PR #565, opened 2025-12-25
Fix overflow issue for exponential ranges.

HuwCampbell:
Because Doubles are used in the calculation of the exponential function, if the word size is greater than 53 bits then the Double representation may be inexact and might end up out of the bounds.

The simplest fix is to special case on the size parameter when calculating the bound, this allows us to just return the bound and skip any conversions.

### DIFF
--- a/hedgehog/test/Test/Hedgehog/Range.hs
+++ b/hedgehog/test/Test/Hedgehog/Range.hs
@@ -14,7 +14,6 @@ prop_exponential_overflow :: Property
 prop_exponential_overflow = withTests 1 . property $ do
   maxBound === Range.upperBound 99 (Range.exponentialBounded :: Range Word64)
 
-
 tests :: IO Bool
 tests =
   checkParallel $$(discover)


hedgehog/test/Test/Hedgehog/Range.hs
```diff
@@ -0,0 +1,21 @@
...
+
+import           Hedgehog
+import qualified Hedgehog.Range as Range
+
+-- | Ensure exponential range doesn't overflow at the upper bound.
+--
+prop_exponential_overflow :: Property
+prop_exponential_overflow = withTests 1 . property $ do
+  maxBound === Range.upperBound 99 (Range.exponentialBounded :: Range Word64)
+
```
moodmosaic:
```suggestion
```

hedgehog/test/Test/Hedgehog/Range.hs
```diff
@@ -0,0 +1,21 @@
...
+
```
moodmosaic:
```suggestion
```

moodmosaic (approved):
LGTM!

PR #573, opened 2026-06-02
Treat string literals as code when classifying comments

### LINKED ISSUE #570: Strings containing "{-" are treated as the start of a comment block when parsing source code for generating pretty error output
This code:

```
  1 module Main where
  2
  3 import Hedgehog
  4
  5 a = "{-"
  6
  7 main :: IO ()
  8 main = do
  9   check . property $ 1 === 2
 10   return ()
```

gives this output:

```
  ✗ <interactive> failed at test/Main.hs:9:24
    after 1 test.
    shrink path: 1:

    ━━━ Failed (- lhs) (+ rhs) ━━━
    - 1
    + 2
```

If we remove `a = "{-"` on line 5 we get this output:

```
  ✗ <interactive> failed at test/Main.hs:7:24
    after 1 test.
    shrink path: 1:

      ┏━━ test/Main.hs ━━━
    5 ┃ main :: IO ()
    6 ┃ main = do
    7 ┃   check . property $ 1 === 2
      ┃   ^^^^^^^^^^^^^^^^^^^^^^^^^^
      ┃   │ ━━━ Failed (- lhs) (+ rhs) ━━━
      ┃   │ - 1
      ┃   │ + 2
    8 ┃   return ()
```

I would expect both cases to give the second output showing the source code. I suspect this is caused by the [classified](https://hackage-content.haskell.org/package/hedgehog-1.7/docs/src/Hedgehog.Internal.Discovery.html#classified) function treating the string `"{-"` as the start of a comment block, even though it appears inside a string literal.

Possible ways to fix this:

1. Add a string literal check to `classified` function.
2. Rewrite `Hedgehog.Internal.Discovery` using `ghc-lib-parser` (which is what tools like `hlint` and `ormolu` uses to parse Haskell code)

SAY-5:
Fixes #570. `classified` treated a `{-` inside a string literal as the start of a block comment, so any declaration after a line like `a = "{-"` was hidden and the source rendering for failing properties broke. This skips over string literal bodies (honouring backslash escapes) so their contents are always classified as code, and adds a Discovery test that the property after such a line is still found.

### DIFF
--- a/hedgehog/src/Hedgehog/Internal/Discovery.hs
+++ b/hedgehog/src/Hedgehog/Internal/Discovery.hs
@@ -180,6 +180,23 @@ classified =
       x : xs ->
         ok x : string k xs
 
+    -- Consume a multiline string literal body after the opening @"""@, up to
+    -- and including the closing @"""@, honouring backslash escapes. As with
+    -- ordinary strings, its contents are code, so a "{-" inside one must not
+    -- open a comment.
+    multiline k = \case
+      [] ->
+        []
+
+      x@(Pos _ '\\') : y : xs ->
+        ok x : ok y : multiline k xs
+
+      x@(Pos _ '"') : y@(Pos _ '"') : z@(Pos _ '"') : xs ->
+        ok x : ok y : ok z : k xs
+
+      x : xs ->
+        ok x : multiline k xs
+
     loop nesting in_line = \case
       [] ->
         []
@@ -198,6 +215,9 @@ classified =
 
       x@(Pos _ '\'') : y@(Pos _ '\\') : z@(Pos _ '"') : w@(Pos _ '\'') : xs | nesting <= 0 ->
         ok x : ok y : ok z : ok w : loop nesting in_line xs
+
+      x@(Pos _ '"') : y@(Pos _ '"') : z@(Pos _ '"') : xs | nesting <= 0 ->
+        ok x : ok y : ok z : multiline (loop nesting in_line) xs
 
       x@(Pos _ '"') : xs | nesting <= 0 ->
         ok x : string (loop nesting in_line) xs
--- a/hedgehog/src/Hedgehog/Internal/Discovery.hs
+++ b/hedgehog/src/Hedgehog/Internal/Discovery.hs
@@ -164,6 +164,22 @@ classified =
     ko =
       Classified Comment
 
+    -- Consume a string literal body after the opening quote, up to and
+    -- including the closing quote, honouring backslash escapes. Characters
+    -- inside a string are code, so a "{-" within it must not open a comment.
+    string k = \case
+      [] ->
+        []
+
+      x@(Pos _ '\\') : y : xs ->
+        ok x : ok y : string k xs
+
+      x@(Pos _ '"') : xs ->
+        ok x : k xs
+
+      x : xs ->
+        ok x : string k xs
+
     loop nesting in_line = \case
       [] ->
         []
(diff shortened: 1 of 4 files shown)


hedgehog/src/Hedgehog/Internal/Discovery.hs
```diff
@@ -164,6 +164,22 @@ classified =
     ko =
       Classified Comment
 
+    -- Consume a string literal body after the opening quote, up to and
+    -- including the closing quote, honouring backslash escapes. Characters
+    -- inside a string are code, so a "{-" within it must not open a comment.
```
moodmosaic, line 188: Do we handle cases like `quote = '"'` -- I think we don't so this can re-trigger the same declaration-hiding behavior #570 is about (trace included above).

Could we either add minimal `Char`-literal skipping or, at least, document the limitation and add a regression test for it?

A few extra string tests (escaped quote, "--", "{- -}") would also be valuable.

moodmosaic (commented):
Thank you @SAY-5 👍

Could we handle character literals, or at least special-case double-quote character literals, and add a regression test?

HuwCampbell:
We also have multiline strings in GHC now, which is pretty great.

hedgehog/src/Hedgehog/Internal/Discovery.hs
```diff
@@ -164,6 +164,39 @@ classified =
...
+    string k = \case
+      [] ->
+        []
+
+      x@(Pos _ '\\') : y : xs ->
+        ok x : ok y : string k xs
```
moodmosaic, line 198: Is this correct for string gaps? This is valid Haskell:
```hs
s = "abc\
\"
```

hedgehog/src/Hedgehog/Internal/Discovery.hs
```diff
@@ -164,6 +164,39 @@ classified =
...
+    -- Consume a multiline string literal body after the opening @"""@, up to
+    -- and including the closing @"""@, honouring backslash escapes. As with
+    -- ordinary strings, its contents are code, so a "{-" inside one must not
+    -- open a comment.
+    multiline k = \case
+      [] ->
+        []
+
+      x@(Pos _ '\\') : y : xs ->
+        ok x : ok y : multiline k xs
```
moodmosaic, line 219: Same here.

hedgehog/src/Hedgehog/Internal/Discovery.hs
moodmosaic: It'd be nice with a regression test like this to find out:

```hs
prop_string_gap_before_closing_quote_does_not_leave_string_open :: Property
prop_string_gap_before_closing_quote_does_not_leave_string_open =
  withTests 1 . property $
    discovered
      [ "module Test where"
      , ""
      , "s = \"abc\\"
      , "\\\""
      , ""
      , "{-"
      , "prop_hidden = ()"
      , "-}"
      ] === []
```

hedgehog/src/Hedgehog/Internal/Discovery.hs
moodmosaic: And the multiline twin alongside it, since that code path fails independently:

```hs
prop_string_gap_in_multiline_string_does_not_leave_string_open :: Property
prop_string_gap_in_multiline_string_does_not_leave_string_open =
  withTests 1 . property $
    discovered
      [ "module Test where"
      , ""
      , "s ="
      , "  \"\"\"abc\\"
      , "\\\"\"\""
      , ""
      , "{-"
      , "prop_hidden = ()"
      , "-}"
      ] === []
```

PR #576, opened 2026-07-02
Fix Haddock formatting

alicelogos:


moodmosaic:
Thank you @alicelogos!

